PR #12, opened 2019-11-12
[gallery] fix (slideshowView): work on item navigation

yhattav:
I worked on the nextItem, getIdxByScroll, arrows, keyboard nav, accessibility or not...
on top of it there is RTL and slideshowLoop. 
I tried to keep it simple but there are lots of features involved.

I need help to make sure I am not "breaking back" your parts of it - RTL and the loop.

https://jira.wixpress.com/browse/PHOT-1344

### DIFF
--- a/packages/gallery/src/components/gallery/slideshowView.js
+++ b/packages/gallery/src/components/gallery/slideshowView.js
@@ -43,7 +43,7 @@ class SlideshowView extends GalleryComponent {
   }
 
   isScrollStart(isRTL = this.props.styleParams.isRTL) {
-
+    
     if (this.container) {
       const {scrollLeft, scrollWidth, clientWidth} = this.container;
       if (isRTL) {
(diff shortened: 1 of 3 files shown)


packages/gallery/src/components/gallery/slideshowView.js
```diff
@@ -105,54 +111,57 @@ class SlideshowView extends GalleryComponent {
   }
 
   //__________________________________end of slide show loop functions__________________________
-  nextItem(direction, isAutoTrigger, scrollDuration = 400) {
-
+  nextItem({direction, isAutoTrigger, scrollDuration = 400, isKeyboardNavigation = false}) {
+    if((!isKeyboardNavigation || !this.props.styleParams.isAccessible) && this.props.styleParams.isGrid && this.props.styleParams.numberOfImagesPerRow) {
```
yhattav: check if you can know if one item is in focus

PR #33, opened 2020-01-08
Jan8 gallery presets

IzaacAyelin:


packages/gallery/src/components/gallery/index.js
```diff
@@ -34,11 +34,14 @@ export default props => {
     styles: galleryProps.styles
   });
 
+  const {galleryType, galleryLayout} = galleryProps.styles;
   let GalleryComponent = ProGallery;
+
   if (isEligibleForLeanGallery(galleryProps)) {
```
yhattav, line 38: Maybe this should go after the whole thing. 
Currently it looks like if you have a gallery with galleryLayout=3000 you will never be eligable for leanGallery.

PR #36, opened 2020-01-12
Simple Gallery Presets 

guysopher:


packages/gallery/src/components/gallery/presets/propTypes.js
```diff
@@ -0,0 +1,76 @@
...
+export const layoutPropTypes = layoutProps => {
+  return {
+    ...basePropTypes,
+    ...layoutProps.reduce((resPropTypes, prop) => resPropTypes[prop] = baseLayoutPropTypes[prop], {})
+  }
+}
+
+export const propsToStyles = props => {
+  return {
+    ...props.styles,
```
yhattav: Move to bottom. protect from unwanted use. if users called something by name it will be the strongest.

PR #47, opened 2020-01-30
change title show to opacity and not visibility

yhattav:


yhattav:
https://jira.wixpress.com/browse/PHOT-1754

PR #62, opened 2020-02-24
Feb21 gallery sp gallery slider image ratio

IzaacAyelin:


packages/gallery/src/components/gallery/presets/gridGallery.js
```diff
@@ -29,9 +29,14 @@ export const fixedStyles = {
 }
 
 export const createStyles = styles => {
+  const processedStyles = {};
+  if (typeof styles.galleryImageRatioFromWix !== 'undefined') {
```
IzaacAyelin: should not be in open source
move this logic to the style builder and pass the relevant cubeRatio according to the layout

PR #69, opened 2020-02-24
Feb19 gallery sp group size

IzaacAyelin:


packages/gallery/src/settings/options/groupSize.js
```diff
@@ -0,0 +1,12 @@
+import { GALLERY_CONSTS, INPUT_TYPES } from '../utils/constants';
+
+export default {
+  title: 'Max Group Size',
+  isRelevant: (styleParams)  => styleParams.collageDensity > 0 && styleParams.groupsPerStrip > 1,
```
IzaacAyelin: fix isRelevant 
should be "always"

PR #77, opened 2020-03-01
Feb27 gallery sp has thumbnails

IzaacAyelin:


packages/gallery/src/settings/index.js
```diff
@@ -70,4 +71,5 @@ export default {
     imageInfoType,
     groupSize,
     collageDensity,
+    hasThumbnails
```
IzaacAyelin: add to sections in playground

PR #81, opened 2020-03-04
March3 gallery sp arrow size

IzaacAyelin:


packages/gallery/src/settings/options/arrowsSize.js
```diff
@@ -0,0 +1,12 @@
+import { GALLERY_CONSTS, INPUT_TYPES } from '../utils/constants';
+
+export default {
+  title: 'Navigation Arrows Size',
+  isRelevant: (styleParams)  => styleParams.oneRow,
```
IzaacAyelin: Maybe showArrows should be relevant here

IzaacAyelin: should be relevant

PR #89, opened 2020-03-05
[gallery] arrowsPosition: tests

IzaacAyelin:


packages/gallery/src/settings/options/arrowsPosition.js
```diff
@@ -0,0 +1,13 @@
+import { GALLERY_CONSTS, INPUT_TYPES } from '../utils/constants';
+import { createOptions } from '../utils/utils';
+
+export default {
+  title: 'Navigation Arrows Position',
+  isRelevant: (styleParams)  => styleParams.oneRow,
```
IzaacAyelin: showArrows should be relevant?

IzaacAyelin: showArrows should be relevant

PR #90, opened 2020-03-07
[gallery] cropOnlyFill

IzaacAyelin:


packages/gallery/src/settings/options/cropOnlyFill.js
```diff
@@ -0,0 +1,10 @@
+import { GALLERY_CONSTS, INPUT_TYPES } from '../utils/constants';
+
+export default {
+  title: 'Crop Only Fill? (find a better name)',
+  description: `When using cubeType (FIT), this option let you keep the image original ratio while not creating
+  margins between the image and the image container.`,
```
IzaacAyelin, line 6: Should be removed from the gallery (cancels crop)

PR #94, opened 2020-03-08
March4 gallery sp groups per row

IzaacAyelin:


packages/gallery/src/settings/options/groupsPerStrip.js
```diff
@@ -0,0 +1,12 @@
+import { GALLERY_CONSTS, INPUT_TYPES } from '../utils/constants';
+
+export default {
+  title: 'Groups per Row',
+  isRelevant: (styleParams)  => !styleParams.oneRow,
```
IzaacAyelin: !isVertical

packages/gallery/src/settings/options/groupsPerStrip.js
```diff
@@ -0,0 +1,12 @@
...
+  type: INPUT_TYPES.NUMBER,
```
IzaacAyelin, line 6: move to wrapper

PR #95, opened 2020-03-08
[gallery] fixedColimns

IzaacAyelin:


packages/gallery/src/settings/options/fixedColumns.js
```diff
@@ -0,0 +1,10 @@
+import { INPUT_TYPES } from '../utils/constants';
+
+export default {
+  title: 'Fixed Columns',
```
IzaacAyelin: Number of columns

packages/gallery/src/settings/options/fixedColumns.js
```diff
@@ -0,0 +1,10 @@
...
+  description: `Set a fixed number of columns in the gallery. Note that this option relies on the options isVertical (set to "true")
```
IzaacAyelin, line 5: fixedColumns should replace imagesPerRow

PR #114, opened 2020-04-01
[layouter] fixed: safeGetItem function

IzaacAyelin:


packages/gallery/src/components/helpers/itemsHelper.js
```diff
@@ -76,13 +76,20 @@ export class ItemsHelper {
...
+    
     for (let c = 0; c < galleryStructure.columns.length; c++) {
       const column = galleryStructure.columns[c];
       const existingColumn = existingStructure.columns[c] || column;
       if (!existingColumn.galleryGroups) {
         existingColumn.galleryGroups = [];
       }
-      const groups = column.groups || column;
+      // remove last group so it will be rebuilt
+      existingColumn.galleryGroups.splice(-1);
```
IzaacAyelin: move to before loop, use. pop()

PR #120, opened 2020-04-22
lastgroup 

meir-galilsoftware:
pop() from empty array is undefined


guysopher:
@meir-galilsoftware - can you specify the scenario in which this fix in required?

meir-galilsoftware:
@guysopher when having multiple views with filter, and the number of images changes between views, the pro-gallery fails to switch correctly between the views. 
for example: switch between tabs in this site:
https://lihie10.wixsite.com/sg270420/profile/lihie/gallery-dev

guysopher:
@meir-galilsoftware 
The tab switching seems to work perfectly.
The thing is that the code is breaking because the data is wrong. If we'll protect this part, it will break somewhere else. We need to figure out the root cause.

PR #124, opened 2020-05-12
Render hover only after interaction

guysopher:


guysopher:
@IzaacAyelin can you review and deploy this?
The change should avoid rendering the Hover element unless it is displayed.
I checked it and it looks good, but tests are failing (since the element is not there)

PR #130, opened 2020-06-01
Remove inline style tags

guysopher:


guysopher:
This branch will break SSR, should be merged after specs.pro-gallery.useCustomElement is used

PR #147, opened 2020-07-05
[gallery] Enable LeanGallery with blueprints

ido-postelnik:


packages/gallery/src/components/gallery/index.js
```diff
@@ -76,8 +76,10 @@ export default class BaseGallery extends React.Component {
         }
       }
     }
+    else if(galleryType === undefined || galleryLayout === LAYOUTS.GRID) {
```
guysopher: Doesn't this break the Blueprints' Grid gallery flow?
Maybe this condition should be specific to the lean gallery

packages/gallery/src/components/gallery/leanGallery/leanGallery.scss
```diff
@@ -1,10 +1,11 @@
 :global {
   div.lean-gallery-gallery {
     display: grid;
+    position: static;
```
guysopher, line 4: why?

packages/gallery/src/components/gallery/index.js
guysopher: You should talk about it with Yonatan, ask him who to render the lean gallery in the blueprints flow

ido-postelnik: I agree it is not too elegant to have this condition twice (in the switch-case and here) but we somehow need to access to the "GridGallery" component. 

By default the GalleryComponent is "ProGallery",
and if the bluePrints aren't defined we won't access the switch-case at all --> so no chance to use the "GridGallery".

packages/gallery/src/components/gallery/leanGallery/leanGallery.scss
ido-postelnik: Here it is just a small discussion I wanted to talk to you.

This change is not mandatory, 
but when using "display: grid" mechanism we dont need to use the "position" property at all,
and because we still use the classes of pro-gallery, I wanted to highlight that we can override the position property to its default value.

packages/gallery/src/components/gallery/index.js
```diff
@@ -7,6 +7,7 @@ import dimensionsHelper from '../helpers/dimensionsHelper';
 import defaultStyles from '../../common/defaultStyles';
 import LAZY_LOAD from '../../common/constants/lazyLoad';
 import utils from '../../common/utils';
+import Blueprints from '../blueprints/Blueprints'
```
guysopher: Check if this import is needed

packages/gallery/src/components/gallery/index.js
guysopher: use the isEligible method - and solve the lean gallery case only
`if (isEligibleForLeanGallery(items, styles)) { <LeanGallery/>}`

guysopher: use `addLayoutStyles` and `addPresetStyles`

guysopher:
```
else if (isEligibleForLeanGallery(items, styles)) {
  galleryProps.styles = addLayoutStyles(addPresetStyles(galleryProps.styles));
  GalleryComponent = LeanGallery;
}
```

guysopher: This can run inside the LeanGallery
`addLayoutStyles(addPresetStyles`

guysopher: styles = {...defaultStyles, ...styles}
      formattedStyles = processLayouts(addPresetStyles(styles));

packages/gallery/src/components/gallery/index.js
```diff
@@ -75,8 +78,11 @@ export default class BaseGallery extends React.Component {
         }
       }
     }
+    else if(isEligibleForLeanGallery(createStylesFromProps(galleryProps))) {
```
guysopher: isEligibleForLeanGallery({...galleryProps, styles: createStyles(galleryProps.styles)})

packages/gallery/src/components/gallery/presets/gridGallery.js
```diff
@@ -35,6 +35,10 @@ export const createStyles = styles => {
   }
 }
 
+export const createStylesFromProps = props => {
```
guysopher: no need for this

packages/gallery/src/components/gallery/leanGallery/leanGallery.js
```diff
@@ -22,16 +25,29 @@ const get = (item, attr) => {
...
-    }
-  }
-
 }
-export default class LeanGallery extends React.Component {
 
+export const formatLeanGalleryStyles = (styles) => {
+	const formattedStyles = processLayouts(addPresetStyles(styles)); // TODO make sure the processLayouts is up to date. delete addLayoutStyles from layoutsHelper when done with it...
+
+	const selectedLayoutVars = [
```
guysopher: no need for that

packages/gallery/src/components/gallery/index.js
```diff
@@ -75,8 +78,11 @@ export default class BaseGallery extends React.Component {
...
+      galleryProps.styles = formatLeanGalleryStyles(galleryProps.styles);
+      GalleryComponent = LeanGallery;
+    }
```
guysopher, line 84: galleryProps.useBlueprints = false

TO CHECK WITH YONATAN

packages/gallery/src/components/gallery/index.js
```diff
@@ -75,8 +78,11 @@ export default class BaseGallery extends React.Component {
         }
       }
     }
+    else if(isEligibleForLeanGallery({...galleryProps, styles: createStyles(galleryProps.styles)})) {
```
guysopher, line 81: the isEligible should also use the `formatLeanGalleryStyles`

PR #153, opened 2020-07-14
Jun2 resize ssr with master

guysopher:


packages/gallery/src/common/utils/index.js
```diff
@@ -283,8 +282,9 @@ class Utils {
     });
   }
 
+  //TODO : Replace with isPrerender mode
```
guysopher, line 285: should we do the todo?
(no ssr references in the core)

packages/gallery/src/components/gallery/index.js
```diff
@@ -7,6 +7,7 @@ import dimensionsHelper from '../helpers/dimensionsHelper';
 import defaultStyles from '../../common/defaultStyles';
 import LAZY_LOAD from '../../common/constants/lazyLoad';
 import utils from '../../common/utils';
+import Blueprints from '../blueprints/Blueprints'
```
guysopher, line 10: this should not be here

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -295,7 +295,7 @@ class SlideshowView extends GalleryComponent {
   }
 
   autoScrollToNextItem = () => {
-    if (!isEditMode() && (isGalleryInViewport(this.props.container) || isPreviewMode())) {
+    if (!isEditMode() && (isGalleryInViewport(this.props.container))) {
```
guysopher, line 298: is this safe to remove?
What happens in preview?

packages/gallery/src/components/item/imageItem.js
```diff
@@ -67,58 +95,59 @@ export default class ImageItem extends GalleryComponent {
...
+              {...preloadProps}
+            />
+            break;
+        }
+
+        imagesComponents.push(preload);
+      }
 
-      return [preload, highres]
+      const shouldRenderHighResImages = !isPrerenderMode() && !utils.isSSR();
```
guysopher, line 133: why are we using isSSR?
We should use componentDidMount to allow highres images

packages/gallery/src/components/styles/animations.scss
```diff
@@ -0,0 +1,5 @@
+:global {
+    .animation-slide {
+        transition: all 0.4s ease
```
guysopher, line 3: this is very bad css.
Use the actual properties you need to animate - never use `all`

packages/ssr-simulator/src/components/gallery/gallery.js
```diff
@@ -3,28 +3,100 @@ import { ProGallery } from 'pro-gallery';
...
-    const items = utils.mixAndSlice(testItems, 50, styles.seed || 1);
+    this.handleResize = this.handleResize.bind(this);
+
+    this.state = {
+      knownContainer: container !== UNKNOWN_CONTAINER,
+      container: container
+    };
+  }
+
+  initContainer(props) {
```
guysopher, line 27: it is actually - `getContainerFromUrl`

packages/gallery/src/common/utils/index.js
aviranhuga: after santa-wrapper will support PRERENDER mode

packages/gallery/src/components/gallery/index.js
aviranhuga: deleted

packages/gallery/src/components/item/imageItem.js
aviranhuga: backward compatibility, we will remove it when Santa will use prerender mode

PR #159, opened 2020-07-19
cssScrollHelper will be in use only when needed (!isSEOMode && (isEditMode() || gotFirstScrollEvent || showMoreClickedAtLeastOnce))

ritabazak:


packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -407,32 +403,27 @@ export class GalleryContainer extends React.Component {
...
         styleParams,
         domId,
-        allowPreloading,
       });
     }
-    return (scrollCss && scrollCss.length > 0) ? scrollCss : this.scrollCss;
+    if (scrollCss && scrollCss.length > 0) {
+      if (JSON.stringify(this.state.scrollCss) !== JSON.stringify(scrollCss)) {
+        this.setState({
+          scrollCss
```
guysopher: why do you need to store it on the state?
The two parameters that trigger it are on the state
Saving too much information on the state is slow

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -450,6 +441,11 @@ export class GalleryContainer extends React.Component {
         },
         () => {
           this.handleNewGalleryStructure();
+          this.getScrollCssIfNeeded({
```
guysopher: why did you move it here?

guysopher (changes requested):

PR #167, opened 2020-07-27
[gallery] leanGallery: calc number of columns - IN PROGRESS

ido-postelnik:


packages/gallery/src/components/gallery/leanGallery/leanGallery.js
```diff
@@ -94,18 +94,46 @@ export default class LeanGallery extends React.Component {
...
 
     return {
       gridTemplateColumns,
       gridGap: `${imageMargin}px`
     };
   }
 
+  calcNumberOfColumns() {
+    // debugger;
+    const { galleryWidth } = this.props.container;
```
guysopher: start with this:
``` 
if (gridStyle === GALLERY_CONSTS.gridStyle.SET_ITEMS_PER_ROW) {
return 0
}
```

guysopher (changes requested):

packages/gallery/src/components/gallery/leanGallery/leanGallery.js
```diff
@@ -94,18 +94,43 @@ export default class LeanGallery extends React.Component {
...
+    const numberOfColumns = gridStyle === GALLERY_CONSTS.gridStyle.SET_ITEMS_PER_ROW ? numberOfImagesPerRow : this.calcNumberOfColumns();
+    const gridTemplateColumns = numberOfColumns > 0 ? `repeat(${numberOfColumns}, 1fr)` : `repeat(auto-fit, minmax(${itemSize}px, 1fr))`;
 
     return {
       gridTemplateColumns,
       gridGap: `${imageMargin}px`
     };
   }
 
+  calcNumberOfColumns() {
```
guysopher: this function should only happen after componentDidMount. Otherwise, the galleryWidth is meaningless.
Call it on componentDidMount and store the value in the state. later, in createGalleryStyles, use that state value (if greater than 0)

packages/gallery/src/components/gallery/leanGallery/leanGallery.js
```diff
@@ -94,18 +94,43 @@ export default class LeanGallery extends React.Component {
     const { styles } = this.props;
     const { gridStyle, numberOfImagesPerRow, imageMargin } = styles;
 
-    // const minmaxFix = 0.75; //this fix is meant to compensate for the css grid ability to use the number as a minimum only (the pro-gallery is trying to get as close as possible to this number)
-    const minmaxFix = 1;
+    const minmaxFix = 1; //this fix is meant to compensate for the css grid ability to use the number as a minimum only (the pro-gallery is trying to get as close as possible to this number)
```
guysopher: if it is 1, remove it.
We don't store old code as comments, github is our archive

guysopher (changes requested):

packages/gallery/src/components/gallery/leanGallery/leanGallery.js
```diff
@@ -40,6 +41,20 @@ export default class LeanGallery extends React.Component {
...
   componentDidMount() {
     this.eventsListener(GALLERY_CONSTS.events.APP_LOADED, {});
+    this.measureIfNeeded();
+    this.setState({
+      numberOfColumns: this.calcNumberOfColumns(this.props)
+    });
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.measureIfNeeded();
```
guysopher: this should be under the same condition:
`(this.props.container.galleryWidth !== nextProps.container.galleryWidth)`

PR #168, opened 2020-07-28
remove unknown dimensions

aviranhuga:


packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -390,20 +390,18 @@ export class GalleryContainer extends React.Component {
     });
   }
 
-  createCssLayoutsIfNeeded(layoutParams, isApproximateWidth = false) {
+  createCssLayoutsIfNeeded(layoutParams) {
```
guysopher, line 393: Wasn't this commented out in the blueprints flow?

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -519,6 +517,12 @@ export class GalleryContainer extends React.Component {
     return can;
   }
 
+  allowSSROpacity() {
+    const {settings = {}} = this.props;
+    const {disableSSROpacity = false} = settings;
+    return !!utils.isSSR() && isPrerenderMode() && !disableSSROpacity;
```
guysopher, line 523: didn't we remove the isSSR in the core?
the prerenderMode is the only condition.

Also, why is the `settings.disableSSRopacity` needed?

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -536,6 +540,12 @@ export class GalleryContainer extends React.Component {
     }
 
     const displayShowMore = this.containerInfiniteGrowthDirection() === 'none';
+    const ssrDisableTransition =
```
guysopher, line 543: Let's discuss this again.
The hydrate should also no receive these transitions.
I think they should be removed completely

packages/gallery/src/components/item/imageItem.js
```diff
@@ -131,7 +131,7 @@ export default class ImageItem extends GalleryComponent {
               ...restOfDimensions,
               backgroundSize: '0.3px',
               backgroundRepeat: 'repeat',
-              transition: 'all 3s ease-in-out'
+              transition: 'all 1s ease-in-out'
```
guysopher, line 134: this is bad, either specify the exact properties that transition or remove this

guysopher (changes requested):

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
yhattav: W/o this there is no this.layoutCss and nothing renders. If this should be removed how will it work?

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
aviranhuga: disableSSRopacity is needed for testing the SSR - when using ?ssrWarmupOnly=true the gallery will load with opacity 0 (if no custom element - thunderbolt), so we need to add ?ssrWarmupOnly=true&disableSSRopacity=true opacity will be 1.

aviranhuga: the opacity is needed only on SSR (not on prerenderMode), maybe I can use setting.enableSsrOpacity from the wrapper and (instead of disableSSRopacity)

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
aviranhuga: i will remove it

packages/gallery/src/components/item/imageItem.js
aviranhuga: i will remove it

PR #171, opened 2020-07-30
rename gallery size to target item size

guysopher:


packages/lib/src/core/presets/thumbnailGallery.js
```diff
@@ -35,6 +35,7 @@ export const createStyles = styles => {
   return {
     ...styles,
     ...fixedStyles,
+    targetItemSize: styles.targetItemSize || styles.gallerySize,
```
yhattav, line 38: Are these styles.gallerySize here so it wont break?
Otherwise if we decide its a breaking change of a rename we could have 0 gallerySize left in the code

ritabazak (commented):
package-lock.json should be removed

packages/lib/src/core/helpers/layoutHelper.js
```diff
@@ -244,21 +244,8 @@ function processLayouts(styles, customExternalInfoRendererExists) {
   //   usm: {}
   // };
 
-  // Backwards compatibility for masonry layout
```
ritabazak, line 247: I'm not really sure, but this is not needed anymore?

packages/lib/src/core/presets/thumbnailGallery.js
guysopher: No - this is the place where the translation from `gallerySize` to `targetItemSize` is happening.
In this specific case, the translation is 1:1.
The translation from `gallerySizePx` and `gallerySizeRatio` to `targetItemSize` is happening in `processLayouts`

PR #179, opened 2020-08-19
[gallery] return items instead of visible items when there's currentIdx

IzaacAyelin:


guysopher (approved):

packages/gallery/src/components/gallery/proGallery/galleryContainerNew.js
```diff
@@ -96,7 +96,7 @@ export class GalleryContainer extends React.Component {
     const { gotFirstScrollEvent } = this.state;
     const scrollY = window.scrollY;
     const {galleryHeight, scrollBase, galleryWidth} = container;
-    if(isSEOMode() || utils.isSSR() || gotFirstScrollEvent || !isSiteMode() || scrollY > 0) {
+    if(isSEOMode() || utils.isSSR() || gotFirstScrollEvent || !isSiteMode() || scrollY > 0 || this.props.currentIdx > 0) {
```
IzaacAyelin, line 99: slideshow renders at first 2 items, which doesn't allow to get to items at a higher index

PR #180, opened 2020-08-19
Aug19 get more items bug slider

IzaacAyelin:


packages/gallery/src/components/gallery/proGallery/galleryContainerNew.js
```diff
@@ -149,6 +148,7 @@ export class GalleryContainer extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    this.getMoreItemsIfNeeded(0);
```
IzaacAyelin, line 152: call getMoreItemsIfNeeded in case of when on new props this way we have the real container after the first run with the defaults in case of unknown dimensions

packages/gallery/src/components/gallery/proGallery/galleryContainerNew.js
```diff
@@ -129,7 +129,6 @@ export class GalleryContainer extends React.Component {
     this.scrollToItem(this.props.currentIdx, false, true, 0);
     this.handleNewGalleryStructure();
     this.eventsListener(GALLERY_CONSTS.events.APP_LOADED, {});
-    this.getMoreItemsIfNeeded(0);
```
guysopher: do not remove this from here. It should appear in both cases (shouldReceiveProps doesn't happen on mount)

guysopher (changes requested):

PR #182, opened 2020-08-22
Aug15 custom slide show sidebar

IzaacAyelin:


packages/playground/src/constants/settings.js
```diff
@@ -92,6 +92,8 @@ export const stylesBySection = {
         'titlePlacement',
         'overlayBackground',
         'allowSlideshowCounter',
+        'slideshowInfoPlacement',
+        'slideshowInfoWidth',
```
yhattav: dont we already have slidehow info height? why not make it the same styleParam?
It will only ever have one of them. so it can be slidehowInfoSize

packages/lib/src/core/helpers/dimensionsHelper.js
```diff
@@ -56,7 +57,11 @@ class DimensionsHelper {
             break;
         }
       } else if (this.styles.isSlideshow) {
```
IzaacAyelin, line 59: place height condition inside if()

packages/lib/src/core/helpers/dimensionsHelper.js
```diff
@@ -66,8 +71,13 @@ class DimensionsHelper {
...
     return this.getOrPutInCache('galleryWidth', () => {
       let width = Math.floor(this.container.width) + (this.getDimensionFix() * 2); //add margins to width and then remove them in css negative margins
-      if (this.styles.arrowsPosition && this.styles.oneRow) {
-        width -= 2 * (this.styles.arrowsSize + 40 + this.styles.imageMargin);
+      if ( this.styles.oneRow) {
+        if (this.styles.arrowsPosition) {
+          width -= 2 * (this.styles.arrowsSize + 40 + this.styles.imageMargin);
+        }
+        if (this.styles.slideshowInfoPlacement === 'ON_THE_RIGHT') {
+          width -= this.styles.slideshowInfoSize;
```
IzaacAyelin, line 79: check with link with santa-wrapper/statics

PR #206, opened 2020-10-20
layout fixer blink

guysopher:


packages/gallery/src/components/gallery/layoutingIndex.js
```diff
@@ -45,9 +45,24 @@ export default class BaseGallery extends React.Component {
       GalleryComponent = LeanGallery;
     } else {
       if (!this.state.blueprint || Object.keys(this.state.blueprint).length === 0){
-        return null;
+        try {
```
guysopher: This is where the blueprint is read from the layoutFixer

packages/lib/src/core/galleryItem.js
```diff
@@ -215,9 +215,9 @@ class GalleryItem {
       resizeUrl(
         this,
         imgUrl,
-        resizeMethod,
+        (this.cubeImages ? resizeMethod : RESIZE_METHODS.FIT),
```
guysopher, line 218: This was a bug.
The low res image must be exactly the same in all layouts - improves cache and rendering

packages/gallery/src/components/gallery/blueprintsIndex.js
```diff
@@ -16,8 +16,22 @@ export default class BaseGallery extends React.Component {
     let galleryProps = { ...otherProps, styles: _styles, eventsListener: _eventsListener, domId};
 
 
-    if(this.props.useBlueprints) {
-      //
+    if (this.props.useBlueprints) {
+      try {
+        const {layoutFixer} = window;
```
yhattav: place this const under the typeof window !== 'undefined' if you are checking it otherwise it fails before the if (IMO the if can be split).

packages/gallery/src/components/gallery/blueprintsIndex.js
```diff
@@ -16,8 +16,22 @@ export default class BaseGallery extends React.Component {
...
+        if (typeof window !== 'undefined' && layoutFixer && !layoutFixer.hydrated) {
+          galleryProps.structure = layoutFixer.structure;
+          galleryProps.container = layoutFixer.container;
+          galleryProps.items = layoutFixer.items;
```
yhattav: if this is not redundant we have a problem in hydrate production.
items and styles should be the same as in SSR. container and structure are the "fixed" props.

packages/gallery/src/components/gallery/blueprintsIndex.js
```diff
@@ -16,8 +16,22 @@ export default class BaseGallery extends React.Component {
...
+          layoutFixer.hydrated = true;
+          console.log('[LAYOUT FIXER] used structure and container from layoutFixer', layoutFixer);
+        } else {
+          return null;
```
yhattav: This will block any BP gallery that doesn't use layoutFixer (although it is a feature, not a must), even after wrapper measurements are 100% good.

packages/gallery/src/components/gallery/layoutingIndex.js
```diff
@@ -45,9 +45,24 @@ export default class BaseGallery extends React.Component {
...
+          const {layoutFixer} = window;
+          if (typeof window !== 'undefined' && layoutFixer && !layoutFixer.hydrated) {
```
yhattav: why not use this as the initial state in the constructor (if there is a layoutFixer, state.blueprints... = layoutFixer...
Otherwise you are in a race wit the onNewProps in the constructor. This should replace it so an initial state wont be set with the first props that arrive.

constructor {
if (layoutFixer) {
state.blueprint = fixer 
} else {
onNewProps(props)
}
}

This will also not put layoutFixer logic in the "canRender".

If you exit the constructor w/o a layoutFixer blueprint you will return null until the state shows a valid blueprint.

packages/gallery/src/components/gallery/proGallery/galleryContainerExtraNew.js
```diff
@@ -48,7 +48,7 @@ export class GalleryContainer extends React.Component {
         this.initialGalleryState = galleryState;
       }
     } catch (_e) {
-      console.warn(_e);
+      console.warn('Cannot create initial state from props (blueprints)', _e);
```
yhattav, line 52: Thanks, I dont want the view components to have the word blueprint. the layoutingIndex should be the only one and it is {...blueprint} to the core.

packages/playground/src/utils/itemResizer.js
```diff
@@ -79,6 +79,8 @@ const resizeUrlImp = (
       options.filters = {
         blur: sharpParams.blur,
       };
+      //this is a hack to avoid using the imageClientSdk for blurry images. These have to stay EXACTLY the same between SSR and CSR
+      return `${item.mediaUrl}/v1/fit/w_250,h_250,al_c,q_30,blur_30/${item.id}.jpg`;
```
yhattav, line 83: This will force this url on any blur. Why not create a function on the galleryItem for this where the rest are.

createUrl('largeThumbnail','blur')  something like that... I dont remember the actual enum naming and options.

Then you use this in places where you want to get this type of constant representation of a mediaUrl.

packages/ssr-simulator/src/components/gallery/gallery.js
```diff
@@ -112,20 +112,20 @@ export default class Gallery extends React.PureComponent {
     //   import ('./layoutFixer').then(console.warn);
     // }
 
+    const Gallery = (this.isSSR() || !urlParams.useBlueprints) ? ProGallery : LayoutingProGallery
```
yhattav, line 114: can you change the ProGallery to blueprintsProGalery? (at the end blueprints/layouting will be the indexes that are left in v3 - in some other naming we choose)

packages/gallery/src/components/gallery/blueprintsIndex.js
guysopher: Won't it save fetching the items from the db?

packages/ssr-simulator/src/components/gallery/gallery.js
guysopher: The layouting is not blueprint?
The ProGallery if when we don't use blueprints

packages/playground/src/utils/itemResizer.js
guysopher: Yes, we should do that...
I don't want to add it here though

packages/gallery/src/components/gallery/blueprintsIndex.js
yhattav: We get the items in the worker anyway. and it should be the exact same amount of items that we get in the ssr

packages/ssr-simulator/src/components/gallery/gallery.js
yhattav: Both use blueprint:

layoutingIndex = same api as before, we do the blueprint for you (bigger bundle that includes layouter etc).

blueprintsIndex = you should provide your deconstructed blueprint in the props. (you plan, we build).

PR #219, opened 2020-11-20
[playground] remove double item number

yhattav:


yhattav:
If you want to let customTitles have a # that will be replaced. otherwise we need to remove the .replace instead

PR #251, opened 2020-12-28
disabaled styles when itemClick set to nothing

yakovhaiilo:


packages/gallery/src/components/item/itemView.js
```diff
@@ -626,7 +626,8 @@ class ItemView extends GalleryComponent {
             placement,
```
IzaacAyelin: maybe we should use an object here

IzaacAyelin (approved):

IzaacAyelin (approved):

PR #274, opened 2021-02-07
Fix Pass customImageRenderer to ViewComponent

maximgeerinck:
Passing `customImageRenderer` like indicated in the docs doesn't seem to work, after passing it to the `ViewComponent` it works

Usage example:

```javascript
<ProGallery
  items={items}
  options={options}
  container={{
    width,
    height,
  }}
  eventsListener={eventsListener}
  scrollingElement={scrollingElement}
  customImageRenderer={customImageRenderer}
/>
  ```

### DIFF
--- a/packages/gallery/package.json
+++ b/packages/gallery/package.json
@@ -2,7 +2,7 @@
   "publishScoped": false,
   "private": false,
   "name": "pro-gallery",
-  "version": "2.4.17",
+  "version": "2.4.18",
   "sideEffects": [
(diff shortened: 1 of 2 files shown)


yhattav:
@guysopher can you take a look. 
Notice - do not merge the version change in the package.json. not sure why its in the changed files.

guysopher:
The fix is only relevant for version 2 (it is adding it on
GalleryContainerNew).
I think we should recommend him to upgrade to version 3

PR #279, opened 2021-02-23
[gallery] prevent animation scroll when not needed

IzaacAyelin:


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -597,6 +607,9 @@ export class GalleryContainer extends React.Component {
       this.currentHoverChangeEvent.currentHoverIdx = eventData;
       window.dispatchEvent(this.currentHoverChangeEvent);
     }
+    if (eventName === GALLERY_CONSTS.events.CURRENT_ITEM_CHANGED) {
+      this.setCurrentSlideshowViewIdx(event.idx);
```
yhattav: No way you think this is ok. `event` doesnt have an idx. you cant pass the item as event because event is not an item. 
you cant just pass things as a third argument and use the third argument for your case. the item needs to be part of the eventData. if the CURRENT ITEM CHANGED event data is more complex, so be it. Talk to me about it it cant stay this way.
Other than that (which is not connected to the actual PR) it looks good.

IzaacAyelin: understood, I'll use the eventData argument

PR #285, opened 2021-03-09
addde styleParam and default

IzaacAyelin:


packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -223,9 +223,12 @@ class VideoItem extends GalleryComponent {
           this.props.actions.setItemLoaded();
           this.setState({ ready: true });
         }}
+        controls={this.props.styleParams.videoControls}
```
yhattav: showVideoControls

yhattav (changes requested):
you need to add tests if you add a new styleParam to the gallery.

PR #297, opened 2021-04-26
Apr20 no index seo

IzaacAyelin:


nirnaor (approved):

packages/gallery/tests/drivers/mocks/items.js
```diff
@@ -422,6 +422,80 @@ const images = [
   },
 ];
 
+const itemsWithDirectShareLink = [
+  {
+    directShareLink:
+      'https://static.wixstatic.com/media/8bb438_1b73a6b067b24175bd087e86613bd00c.jpg',
```
nirnaor: Looks like all these objects are duplicated, except for one parameter (the `directShareLink`). Can you make something like this?:
```
const itemsWithDirectShareLink = ['8bb438_1b73a6b067b24175bd087e86613bd00c.jpg', '8bb438_e8735e0102554d6ead385a48e2b9e69a.jpg'].map(createImagesWithDirectShareLink]
```

I think you can reduce the amount of code by 500% that way.
Usually too much mock data makes it difficult to track the difference between the test data

packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    await driver.update();
+    const linkProps = driver.find.selector('a').props();
+    expect(linkProps).to.include({ rel: 'nofollow' });
+  });
+
+  it('should not use "rel" attribute for preventing seo indexing - itemClick = expand', async () => {
+    Object.assign(sampleItemViewProps, {
+      styleParams: {
+        itemClick: GALLERY_CONSTS.itemClick.EXPAND,
+        shouldIndexDirectShareLinkInSEO: true,
```
nirnaor: The difference between the two tests is only the `shouldIndexDirectShareLinkInSEO`. Can you reuse the code?

packages/lib/src/settings/options/shouldIndexDirectShareLinkInSEO.js
```diff
@@ -0,0 +1,14 @@
+import { INPUT_TYPES } from '../utils/constants';
+import { default as GALLERY_CONSTS } from '../../common/constants';
+
+export default {
+  title: 'Search engines indexing for direct share links',
+  description: `Choose whether you want the items with share link to be indexed by crawlers in SEO or not`,
+  isRelevant: (sp) =>
```
nirnaor, line 7: This is just my taste but I cna 
```[GALLERY_CONSTS.itemClick.FULLSCREEN, GALLERY_CONSTS.itemClick.EXPAND].includes(sp.itemClick)
```

nirnaor (changes requested):

packages/lib/src/settings/options/shouldIndexDirectShareLinkInSEO.js
IzaacAyelin: not sure I understand here, I see an empty line

packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js
IzaacAyelin: not sure how. I think it's easier to read it this way (or is it not? what do you think?)

packages/lib/src/settings/options/shouldIndexDirectShareLinkInSEO.js
nirnaor: oops my bad. I would do (but it's just a matter of style)
```
[GALLERY_CONSTS.itemClick.FULLSCREEN, GALLERY_CONSTS.itemClick.EXPAND].includes(sp.itemClick)
```
You gain two things:
1. You don't have a duplication of `sp.itemClick ===`
2. You can easily add more values to the array in the future

IzaacAyelin: cool, I'll do that

packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js
nirnaor: I created an example for you in [this link](https://github.com/wix/pro-gallery/blob/88c814998003be280cefbe530daaa1282711ac1b/packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js#L26).
I think that the reader understands the difference between the tests much better this way. 
This is for you to decide, it's just my preference

nirnaor (approved):
Final comment is [here](https://github.com/wix/pro-gallery/blob/88c814998003be280cefbe530daaa1282711ac1b/packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js#L26)
I'll leave it to you to decide if you want to take it.

packages/gallery/tests/styleParams/shouldIndexDirectShareLinkInSEO.spec.js
IzaacAyelin: cool, seems cleaner. I think I'll go with your approach, thanks

PR #301, opened 2021-05-02
2 may item event

yakovhaiilo:
send all data instead of partial data.

### DIFF
--- a/packages/gallery/src/components/item/itemView.js
+++ b/packages/gallery/src/components/item/itemView.js
@@ -118,17 +118,20 @@ class ItemView extends GalleryComponent {
         this.props.idx
       );
     }
-    this.props.actions.eventsListener(GALLERY_CONSTS.events.ITEM_FOCUSED, {
-      styleParams: this.props.styleParams,
-      itemStyle: this.props.style,
-    });
+    this.props.actions.eventsListener(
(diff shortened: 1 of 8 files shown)


yhattav (approved):
send to nir

nirnaor (approved):

packages/gallery/src/components/item/itemView.js
```diff
@@ -118,17 +118,20 @@ class ItemView extends GalleryComponent {
...
   }
 
   onBlur() {
     if (this.props.styleParams.isAccessible) {
       this.props.actions.eventsListener(GALLERY_CONSTS.events.HOVER_SET, -1);
     }
-    this.props.actions.eventsListener(GALLERY_CONSTS.events.ITEM_LOST_FOCUS);
+    this.props.actions.eventsListener(
+      GALLERY_CONSTS.events.ITEM_LOST_FOCUS,
+      this.props
```
nirnaor, line 133: @yakovhaiilo What is the missing prop that is needed by the event listener?

yakovhaiilo: @nirnaor it's not a missing prop, I'm not using it but our consumers may want to use "ITEM_LOST_FOCUS" event, and it's was decided, that we send all the data when we listening to an item event.

PR #304, opened 2021-05-04
[lib] fix: keep low res images ratio

IzaacAyelin:


nirnaor (changes requested):
`removeLowResImageTimeoutId` will always be undefined therefor it should not stay in the code at all. You can remove it from the [componentWillUnmount](https://github.com/wix/pro-gallery/blob/210cd4cf44e519cb60c6e74f88fe78a3c7eff282/packages/gallery/src/components/item/imageItem.js#L43) and the rest of the code.


Or am I missing something?

IzaacAyelin:
oh forgot to remove the other parts setTimout parts, i'll delete it

nirnaor (approved):

PR #308, opened 2021-05-05
Debug actual source code

nirnaor:
### Why?
We work on open source code so I want to be able to debug it with the same experience that the create-react-app is offering. 
There were two things that bothered me before this PR:

1. When I ran `npm run start` the react-app I had linked to the `pro-gallery` didn't auto refresh on changes.
2. I could only debug the es5 version in the browser, which made it much difficult (screenshot attached)

es5progallery

### What? 

1. Adding a separate configuration file to target `es2017` for dev purposes.
2. Adding `start:dev` which keeps the watch flag at the end of the command

The result is in the gif attached: the react app is auto-refreshing and the source code I debug is the same as the one in my text editor

progallerywatchmodees2017

### DIFF
--- a/packages/gallery/tsconfig-esm-dev.json
+++ b/packages/gallery/tsconfig-esm-dev.json
@@ -1,0 +1,6 @@
+{
+  "extends": "./tsconfig-esm.json",
+  "compilerOptions": {
+    "target": "ES2017"
+  }
+}
--- a/packages/gallery/package.json
+++ b/packages/gallery/package.json
@@ -23,7 +23,7 @@
     "watch": "npm run start",
     "start": "tsc -w && npm run css-watch",
     "build": "tsc && tsc -p tsconfig-esm.json && npm run css-compile",
-    "build:dev": "tsc && tsc -p tsconfig-esm-dev.json && npm run css-compile",
+    "start:dev": "npm run css-compile && tsc && tsc -w -p tsconfig-esm-dev.json",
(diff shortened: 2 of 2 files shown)


packages/gallery/tsconfig-esm-dev.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig-esm.json",
+  "compilerOptions": {
```
yhattav, line 3: Maybe we should copy the existing compilerOptions and adding the es2017 on top.
These are settings I worked on to have an easier TS start. (noImplicitAny etc).
Otherwise the tsc will break the build with different "reasons" than the rest, WDYT?

yhattav: Thanks to Rita, I see the extend :)

yhattav (approved):

packages/gallery/tsconfig-esm-dev.json
nirnaor: I added a separate file and configuration because I didn't want to touch the existing ones. 
I was not sure why the `target` was configured to `es5`. 
If we can change it then this configuration file can be totally removed

ritabazak: Maybe we can change it actually, the tsconfig.json to use ES2017 and not es5.
I'll test it locally and update :)

PR #309, opened 2021-05-05
Images from web

nirnaor:
### Why?
When a user copy pastes code from documentation and it doesn't work out of the box - he is leaving to search for something that works. We will always want our users to start from a **working state**, and then gradually configure and expand their use. But they should always start from the happy flow of a working gallery

### What?

1. Use default items from the web in the playground code generator
2. Do the same on the README.md of the root
3. Remove the video samples beacuse they are dependent on a `resizeMediaUrl` functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,17 +32,83 @@ yarn add pro-gallery
 ```jsx
 import { ProGallery } from 'pro-gallery';
 import 'pro-gallery/dist/statics/main.css';
+export default function Gallery() {
+	// Add your images here...
+	const items = [
+		{ // Image item:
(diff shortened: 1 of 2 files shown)


guysopher (approved):

yhattav (approved):

ritabazak (approved):
I think that we need to make sure that we will re-add an example of a video item in the future.

nirnaor:
As soon as we will change the API to be something that the user can
understand, we will add the video example.
For now I don't think it's usable for a user outside of wix

PR #321, opened 2021-05-12
replace oneRow with scrollDirection

yhattav:
## What?
scrollDirection (enum) will be the StyleParam defining if the gallery should be a vertical or a horizontal one.

## Why?
We currently have two different ways to define a gallery as being horizontal or vertical.
oneRow: boolean (true = horizontal, as being a "one row").
scrollDirection: number ( exported const by the lib).
We want one.

## How?
* Keep scrollDirection. remove oneRow.
* Create a style migrator(styleMigrator) to take oneRow into account when it's passed to the gallery and convert it to scrollDirection.
* stylesMigrator is a function that is exported by the LIB.
* stylesMigrator is connected in the addPresetStyles for all the gallery rendering flows.
* stylesMigrator is also connected in the layouter itself to make sure it also doesn't break (it is a module on it's own).
* A warning message added when stylesMigrator finds a usage of the deprecated oneRow.
* An error message added when the isRelevant of oneRow is used (this is used by settings etc. as in user and not UoU).

## Testing?
* Gallery unitTests will changed to use only scrollDirection (mounting the components).
* e2e tests (integration and SP) will be kept as is to make sure this change doesn't break anything.
* No tests added or changed in their expect().
* We even have a oneRow.e2e that is kept to keep us safe when oneRow is still used.


nirnaor:
@yhattav Notice that there are a lot of files here that contain formatting changes so it will be difficult to go over it. You might wanna pull or cherry pick your commits on latest master.

nirnaor (approved):
@yhattav 
If the tests are now dependent on the new key `scrollDirection`, can't you remove the `oneRow` from the tests and use the `scrollDirection` on the tests?

yhattav:
> @yhattav
> If the tests are now dependent on the new key `scrollDirection`, can't you remove the `oneRow` from the tests and use the `scrollDirection` on the tests?

Yes,
In the gallery tests the oneRow was replaced by scrollDirection.
In the integration e2e tests I didn't replace it. 
some of those were talked as json from our users and seeing they don't break with how they use it is important. If they use oneRow I want to test that it works.

packages/lib/src/core/helpers/stylesMigrator.js
```diff
@@ -0,0 +1,35 @@
+import utils from '../../common/utils/index';
```
nirnaor, line 1: @yhattav 
This will later be inject as an argument? From the wix-private repos?

packages/lib/src/core/helpers/stylesMigrator.js
```diff
@@ -0,0 +1,35 @@
...
+
+    if (utils.isUndefined(styles.scrollDirection)) {
+      styles.scrollDirection = GALLERY_CONSTS.scrollDirection.VERTICAL;
+    }
+
+    // we keep the oneRow as its an API outgoing from the gallery in events
+    // delete styles.oneRow;
+
+    // This code was removed from processStyles:
+    styles.oneRow =
```
nirnaor, line 24: 👑

PR #323, opened 2021-05-18
18 may auto slide pause on focus

yakovhaiilo:
On A11Y mode only :
On horizontal layouts + auto slide-
When the user is in focus mode (a11y) on one of these elements 
Item (video, image, text item) 
Buttons
Navigation Arrows 
then, the auto slide should pause (just like a pause on hover). 

When the user press Esc and the focus go to the gallery container, the auto slide should run again.

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/slideshowView.js
+++ b/packages/gallery/src/components/gallery/proGallery/slideshowView.js
@@ -24,6 +24,8 @@ class SlideshowView extends GalleryComponent {
     this.navigationInHandler = this.navigationInHandler.bind(this);
     this.scrollToThumbnail = this.scrollToThumbnail.bind(this);
     this.stopAutoSlideshow = this.stopAutoSlideshow.bind(this);
+    this.onFocus = this.onFocus.bind(this);
+    this.onBlur = this.onBlur.bind(this);
     this.onAutoSlideShowButtonClick =
       this.onAutoSlideShowButtonClick.bind(this);
     this.startAutoSlideshowIfNeeded =
@@ -46,6 +48,7 @@ class SlideshowView extends GalleryComponent {
       hideLeftArrow: !props.isRTL,
       hideRightArrow: props.isRTL,
       shouldBlockAutoSlideshow: false,
+      isInFocus: false,
     };
     this.lastCurrentItem = undefined;
     this.shouldCreateSlideShowPlayButton = false;
@@ -460,6 +463,11 @@ class SlideshowView extends GalleryComponent {
         e.preventDefault();
         this._next({ direction: 1, isKeyboardNavigation: true });
         return false;
+      case 27: // esc
+      if(this.props.galleryContainerRef) {
+        this.props.galleryContainerRef.focus();
+      }
+      return false;
     }
     return true; //continue handling the original keyboard event
   }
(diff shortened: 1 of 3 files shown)


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1303,27 +1311,44 @@ class SlideshowView extends GalleryComponent {
...
       this.setState({ shouldBlockAutoSlideshow: should }, () => {
-        this.startAutoSlideshowIfNeeded(props.styleParams);
+        this.startAutoSlideshowIfNeeded(styleParams);
       });
     } else {
       return;
     }
   }
 
+  onFocus(){
```
IzaacAyelin, line 1343: onFocus will trigger this element or any child of this element is in focus and that is why we don't need to specifically check if the items, arrows, or buttons are in focus, we just need to be careful with that

yakovhaiilo: 👌

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1303,27 +1311,44 @@ class SlideshowView extends GalleryComponent {
   //-----------------------------------------| REACT |--------------------------------------------//
 
   blockAutoSlideshowIfNeeded(props = this.props) {
-    const { isGalleryInHover } = props;
-    const { pauseAutoSlideshowClicked, shouldBlockAutoSlideshow, isInView } =
+    const { isGalleryInHover, styleParams } = props;
+    const { pauseAutoSlideshowClicked, shouldBlockAutoSlideshow, isInView,isInFocus } =
```
yhattav: not formatted right

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1303,27 +1311,44 @@ class SlideshowView extends GalleryComponent {
...
     let should = false;
     if (!isInView || pauseAutoSlideshowClicked) {
       should = true;
     } else if (
-      isGalleryInHover &&
-      props.styleParams.pauseAutoSlideshowOnHover
+      styleParams.pauseAutoSlideshowOnHover
     ) {
-      should = true;
+      should = isGalleryInHover || isInFocus && styleParams.isAccessible;
```
yhattav: Whats different between SP.isAccessible and SP.pauseAutoSlide...Hover ? (or the others?)
Why is one in an if and the rest in the assignment logic?
I would rather have them all in the if like they were before. if the logic is more specific now it should be there.
if the || is bothering you, this can be 2 else if
first one for the hover logic, and 2nd for the focus logic.
both assigning true to should.

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1303,27 +1311,44 @@ class SlideshowView extends GalleryComponent {
   //-----------------------------------------| REACT |--------------------------------------------//
 
   blockAutoSlideshowIfNeeded(props = this.props) {
-    const { isGalleryInHover } = props;
-    const { pauseAutoSlideshowClicked, shouldBlockAutoSlideshow, isInView } =
+    const { isGalleryInHover, styleParams } = props;
```
yhattav, line 1316: 👍

packages/gallery/src/components/gallery/proGallery/slideshowView.js
yakovhaiilo: SP.pauseAutoSlideOnHover is the condition that should appear in both cases(pauseOnHover && pauseOnFocus), and SP.isAccessible should only appear for the isInFocus state, It will also work the way you mentioned here, it's just my writing preference but I'll change it.

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -460,6 +463,11 @@ class SlideshowView extends GalleryComponent {
         e.preventDefault();
```
nirnaor: I see two problems in this code:

1. This entire logic is duplicated. The only difference between these two `case`s is the `direction`. 
2. There's too much logic here that can be extracted (and tested) outside of the class (no `this` or `bind`).


[This commit](https://github.com/wix/pro-gallery/commit/e828430885db46f41d676a0c014e0807ddb7dcc9#diff-521cda42545179515a78af227d78e504a0391b727945c20b1aa94afa1f621156R464-R472) has the two above issues fixed

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
yakovhaiilo: Thanks👍

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -446,22 +458,18 @@ class SlideshowView extends GalleryComponent {
...
-        this._next({ direction: -1, isKeyboardNavigation: true });
-        return false;
-      case 39: //right
-      case 40: //down
-      case 32: //space
-      case 34: //page down
-        e.preventDefault();
-        this._next({ direction: 1, isKeyboardNavigation: true });
-        return false;
+    const nextKeys = [32, 33, 34, 37, 38, 39, 40]
```
yhattav, line 461: please keep the comments on char code vs key for readability 
like so:
[ 32, // ?
  33, // page up
.....
]

nirnaor (approved):

PR #324, opened 2021-05-19
[gallery] added single item view state and removed placeholder when n…

IzaacAyelin:
…eeded

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -726,6 +726,8 @@ export class GalleryContainer extends React.Component {
         className={this.props.isPrerenderMode ? 'pro-gallery-prerender' : ''}
         onMouseEnter={this.onMouseEnter}
         onMouseLeave={this.onMouseLeave}
+        ref={e => this.galleryContainerRef = e}
+        tabIndex={-1}
       >
         <ScrollIndicator
           domId={this.props.domId}
@@ -768,6 +770,7 @@ export class GalleryContainer extends React.Component {
           proGalleryRegionLabel={this.props.proGalleryRegionLabel}
           firstUserInteractionExecuted={this.state.firstUserInteractionExecuted}
           isGalleryInHover={this.state.isInHover}
+          galleryContainerRef={this.galleryContainerRef}
           actions={{
             ...this.props.actions,
             findNeighborItem: this.findNeighborItem,
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -42,7 +42,6 @@ export class GalleryContainer extends React.Component {
     this.getIsScrollLessGallery = this.getIsScrollLessGallery.bind(this);
     this.onMouseEnter = this.onMouseEnter.bind(this);
     this.onMouseLeave = this.onMouseLeave.bind(this);
-    this.getIsSingleItemHorizontalDisplay = this.getIsSingleItemHorizontalDisplay.bind(this);
     this.videoScrollHelper = new VideoScrollHelperWrapper(
       this.setPlayingIdxState
     );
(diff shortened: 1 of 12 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -350,17 +352,22 @@ export class GalleryContainer extends React.Component {
...
       container,
       structure,
       isScrollLessGallery: this.getIsScrollLessGallery(styles),
+      isSingleItemHorizontalDisplay: this.getIsSingleItemHorizontalDisplay(styles),
     };
     return newState;
   }
-
+  getIsSingleItemHorizontalDisplay(styles) {
+    return styles.oneRow &&
```
yhattav: please use 
scrollDirection === GALLERY_CONSTS.scrollDirection.HORIZONTAL
instead of oneRow

packages/gallery/src/components/item/itemView.js
```diff
@@ -26,6 +26,7 @@ import VideoItemWrapper from './videos/videoItemWrapper';
 class ItemView extends GalleryComponent {
   constructor(props) {
     super(props);
+    console.log('constructor ===> ',props);
```
yhattav: del

packages/gallery/src/components/item/itemView.js
```diff
@@ -1059,6 +1060,7 @@ class ItemView extends GalleryComponent {
   }
 
   componentDidUpdate(prevProps) {
+    console.log('update ===> ',this.props);
```
yhattav: del

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -350,17 +352,22 @@ export class GalleryContainer extends React.Component {
...
       styles,
       container,
       structure,
       isScrollLessGallery: this.getIsScrollLessGallery(styles),
+      isSingleItemHorizontalDisplay: this.getIsSingleItemHorizontalDisplay(styles),
     };
     return newState;
   }
-
+  getIsSingleItemHorizontalDisplay(styles) {
```
IzaacAyelin: should be a function under utils

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -350,7 +350,6 @@ export class GalleryContainer extends React.Component {
 
     this.createCssLayoutsIfNeeded(layoutParams);
     this.createDynamicStyles(styles, isPrerenderMode);
-
```
nirnaor: Formatting changes should not be a part of a PR

packages/lib/src/common/utils/index.js
```diff
@@ -6,6 +6,7 @@ import {
   isPreviewMode,
   isFormFactorMobile,
 } from '../window/viewModeWrapper';
+import GALLERY_CONSTS from '../constants'
```
nirnaor, line 9: I prefer to avoid backward imports (`import x from '../../'`) since it makes the file less portable

packages/gallery/src/components/item/videos/videoItemWrapper.js
```diff
@@ -63,8 +64,11 @@ class VideoItemWrapper extends ImageItem {
       'settings',
       'actions',
     ]);
+    const shouldCreatePlaceHoler =
+      utils.isSingleItemHorizontalDisplay(styleParams) && 
+      styleParams.videoPlay === GALLERY_CONSTS.videoPlay.AUTO;
 
-    return (
+    return shouldCreatePlaceHoler ? null : (
```
nirnaor: How many components are rendering the video item wrapper? Can you pass the shouldCreatePlaceholder (state maybe) to an upper level? And this component will always do one thing(render) ?

nirnaor (approved):
@IzaacAyelin 
I have some questions that don't block the merge

packages/gallery/src/components/item/videos/videoItemWrapper.js
IzaacAyelin: the video item wrapper is rendered on every video in the gallery, I thought it was the best place to put the shouldCreatePlaceholder since it is specific to videos (it should actually be called shouldCreateVideoPlaceholder)

packages/lib/src/common/utils/index.js
IzaacAyelin: is there a way to avoid it?

PR #325, opened 2021-05-19
fix video center alignment when in a tall container

yhattav:
Why?
videos are acting strange in the new possibilities that presets present.
Container ratios can now be set by the preset and not by the ratio of the item.

How?
changed css to 100% height instead of the auto it had. This does it CCS-wise. 
width acts different than height and doesn't need this tweek.


nirnaor (approved):
Is the height of the parent defined?

yhattav:
> Is the height of the parent defined?

it is two up. as pixel css width and height

PR #327, opened 2021-05-23
add keyup event listener to ESC keyCode + remove pauseAutoSlideShowOn…

yakovhaiilo:
accessibility changes:

when visitors hit the Esc key (keyCode === 27), move the focus to the gallery container.
+ pauseOnFocus : remove unnecessary check from blockAutoSlideShow (refactor)


nirnaor (approved):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1326,7 +1326,6 @@ class SlideshowView extends GalleryComponent {
       should = true;
     } else if (
       isInFocus &&
-      styleParams.pauseAutoSlideshowOnHover &&
```
yhattav, line 1329: why is this deleted?

PR #329, opened 2021-05-25
heddle ARIA attributes in core

yakovhaiilo:
accessibility changes.

When the gallery is initialized in "application" mode:

* set role="application" (instead of role="region"),
* set aria-roledescription="gallery application"

When the gallery NOT is initialized in "application" mode:

* set the aria-label attribute value to "feed" (verticals can still override it, e.g. "Blog feed, Select a post to read").
* set the container role to "region"

link to the "photography-pro-gallery" changes: https://github.com/wix-private/photography-pro-gallery/pull/22/files

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -722,6 +722,8 @@ export class GalleryContainer extends React.Component {
       ? SlideshowView
       : GalleryView;
 
+    const getAriaAttributes = this.getAriaAttributes();
+    
     if (utils.isVerbose()) {
       console.count('PROGALLERY [COUNTS] - GalleryContainer (render)');
       console.log(
@@ -783,7 +785,7 @@ export class GalleryContainer extends React.Component {
           firstUserInteractionExecuted={this.state.firstUserInteractionExecuted}
           isGalleryInHover={this.state.isInHover}
           galleryContainerRef={this.galleryContainerRef}
-          getAriaAttributes={this.getAriaAttributes}
+          getAriaAttributes={getAriaAttributes}
           actions={{
             ...this.props.actions,
             findNeighborItem: this.findNeighborItem,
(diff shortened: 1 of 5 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -637,6 +638,17 @@ export class GalleryContainer extends React.Component {
     }
   }
 
+  getAriaAttributes() {
```
nirnaor: Please remove this function outside of the class (and from this file)
Remove the dependency of `this` from the function. 
Accept the relevant props as argument

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -768,10 +782,10 @@ export class GalleryContainer extends React.Component {
           customNavArrowsRenderer={this.props.customNavArrowsRenderer}
           playingVideoIdx={this.state.playingVideoIdx}
           noFollowForSEO={this.props.noFollowForSEO}
-          proGalleryRegionLabel={this.props.proGalleryRegionLabel}
           firstUserInteractionExecuted={this.state.firstUserInteractionExecuted}
           isGalleryInHover={this.state.isInHover}
           galleryContainerRef={this.galleryContainerRef}
+          getAriaAttributes={getAriaAttributes}
```
nirnaor: pass/bind relevant props to the imported method

nirnaor (changes requested):
This class is too long and most of it's logic can be extracted outside of `react`, `this` and it's state. 
Since we are not going to refactor all of it, I want that new functions will be written as **stateless** functions, idependent of `this` and of `react`

nirnaor (approved):

packages/lib/src/common/utils/index.js
```diff
@@ -652,6 +652,17 @@ class Utils {
       styles.cubeImages &&
       styles.cubeRatio === '100%/100%';
   }
+
+  getAriaAttributes(props) {
+    const { proGalleryRole, proGalleryRegionLabel } = props;
```
yhattav: This could be destructed in the arguments that the function gets. so it will expect an object with these keys and we can get rid of "props".  
getAriaAttributes({ proGalleryRole, proGalleryRegionLabel }) {}

yakovhaiilo: 👍

PR #333, opened 2021-05-25
hide high res image until loaded

guysopher:


yhattav:
@IzaacAyelin please review this together with your new CSS for the preload image. This should be a working system

PR #344, opened 2021-06-02
June1 streched images

IzaacAyelin:
**What** - adding a styleParam (strechSmallImages) that allows us to display images in their original size.

**Why** - when a user adds an image that is too small the gallery will stretch the image to fit its container, which results in a pixelated image if the container is too big.

### DIFF
--- a/packages/gallery/src/components/item/itemView.js
+++ b/packages/gallery/src/components/item/itemView.js
@@ -361,7 +361,6 @@ class ItemView extends GalleryComponent {
     let dimensions = {};
     const { height, width, maxWidth, maxHeight } = style;
     const shouldUseMaxDims = maxHeight < height && maxWidth < width && !styleParams.strechSmallImages;
-
     if (!isGridFit) {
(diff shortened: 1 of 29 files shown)


nirnaor (approved):

packages/gallery/src/components/item/itemView.js
```diff
@@ -359,27 +359,39 @@ class ItemView extends GalleryComponent {
     const isGridFit = styleParams.cubeImages && styleParams.cubeType === 'fit';
 
     let dimensions = {};
-
-    if (!isGridFit) {
+    const { height, width, maxWidth, maxHeight } = style;
+    const shouldUseMaxDims = maxHeight < height && maxWidth < width && styleParams.useMaxDimensions;
```
IzaacAyelin: math.min(height, styleParam.useMaxDimensions ? maxHeight : infinity)
math.min(width, styleParam.useMaxDimensions ? maxWidth : infinity)

PR #345, opened 2021-06-03
June2 esaein animations

IzaacAyelin:
**What** - adding a styleparam for the slide animation transition
**why** - we want to allow users to customize the transition by passing a param that will be set in the css


packages/gallery/src/components/helpers/scrollHelper.js
```diff
@@ -217,7 +221,14 @@ function isWithinPaddingHorizontally({
   return res.before < padding && res.after < padding;
 }
 
-function horizontalCssScrollTo(scroller, from, to, duration, isRTL) {
+function horizontalCssScrollTo(
+  scroller,
```
nirnaor, line 223: I have a stupid rule of thumb that I use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring) for functions with more than 3 arguments. You can adopt it if you want

nirnaor (approved):

PR #348, opened 2021-06-05
added out of gallery

yakovhaiilo:
Add a keyUp event listener to the last gallery item. if keyCode === 40, shift the focus to #outofgallery element.

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -37,7 +37,6 @@ export class GalleryContainer extends React.Component {
     this.setPlayingIdxState = this.setPlayingIdxState.bind(this);
     this.getVisibleItems = this.getVisibleItems.bind(this);
     this.findNeighborItem = this.findNeighborItem.bind(this);
-    this.focusGalleryContainer = this.focusGalleryContainer.bind(this);
     this.setCurrentSlideshowViewIdx =
       this.setCurrentSlideshowViewIdx.bind(this);
     this.getIsScrollLessGallery = this.getIsScrollLessGallery.bind(this);
@@ -548,12 +547,6 @@ export class GalleryContainer extends React.Component {
     });
   }
 
-  focusGalleryContainer(){
-    if(this.galleryContainerRef){
-      this.galleryContainerRef.focus();
-    }
-  }
-
   toggleLoadMoreItems() {
     this.eventsListener(
       GALLERY_CONSTS.events.LOAD_MORE_CLICKED,
@@ -734,7 +727,7 @@ export class GalleryContainer extends React.Component {
         className={this.props.isPrerenderMode ? 'pro-gallery-prerender' : ''}
         onMouseEnter={this.onMouseEnter}
         onMouseLeave={this.onMouseLeave}
-        ref={e => this.galleryContainerRef = e}
+        ref={(e) => this.galleryContainerRef = e}
         tabIndex={-1}
       >
         <ScrollIndicator
@@ -779,6 +772,8 @@ export class GalleryContainer extends React.Component {
           proGalleryRole={this.props.proGalleryRole}
           firstUserInteractionExecuted={this.state.firstUserInteractionExecuted}
           isGalleryInHover={this.state.isInHover}
+          galleryContainerRef={this.galleryContainerRef}
+          outOfViewComponent={this.outOfViewComponent}
           actions={{
             ...this.props.actions,
             findNeighborItem: this.findNeighborItem,
@@ -787,7 +782,6 @@ export class GalleryContainer extends React.Component {
             setWixHeight: () => {},
             scrollToItem: this.scrollToItem,
             scrollToGroup: this.scrollToGroup,
-            focusGalleryContainer: this.focusGalleryContainer,
           }}
           {...this.props.gallery}
         />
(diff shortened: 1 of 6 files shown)


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -465,7 +465,19 @@ class SlideshowView extends GalleryComponent {
 
     if (nextKeys.includes(code)) {
       e.preventDefault();
-      this._next({ direction: getDirection(code), isKeyboardNavigation: true });
+      const activeItemIdx = window.document.activeElement.getAttribute('data-idx');
+
+      const shouldOutOfGallery =
```
ritabazak: I think the name of the const should be changed,
check code first,
what if there is no totalItemsCount?

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -814,6 +808,11 @@ export class GalleryContainer extends React.Component {
             <style dangerouslySetInnerHTML={{ __html: this.dynamicStyles }} />
           )}
         </div>
+        {this.props.proGalleryRole === 'application' && (
+          <span ref={(e) => this.outOfViewComponent = e} tabIndex={-1} className="sr-only">
+            out of gallery
```
ritabazak: Need to check what to do about this plain text here

yakovhaiilo: We will add a key

ritabazak: Also please change the className (something like "sr-only-XXXXX") or 2 class names, such as "sr-only XXXXXX"

packages/gallery/src/components/gallery/proGallery/galleryView.js
```diff
@@ -47,6 +47,11 @@ class GalleryView extends GalleryComponent {
           break;
         case 40: //down
           newIdx = findNeighborItem(idx, 'down');
+          if((this.props.totalItemsCount -1) === newIdx && newIdx === this.state.currentIdx){
```
ritabazak, line 50: Why not  to leave only (newIdx === this.state.currentIdx) in the cond? When newIdx === this.state.currentIdx?

yakovhaiilo: Because the requirement that this only happen is that we are on the last item,

yakovhaiilo: If we leave such a condition we can leave the gallery even though we are on the other items in the last row

packages/gallery/src/components/gallery/proGallery/slideshowView.js
yakovhaiilo: if is no totalItemsCount, some of our logic will not work properly because  we also use it in other places,
So I think our consumers must always pass it on,

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -458,22 +458,37 @@ class SlideshowView extends GalleryComponent {
...
       e.preventDefault();
+      const activeItemIdx = window.document.activeElement.getAttribute('data-idx');
+
+      const shouldFocusOutOfViewComponent = activeItemIdx &&
+        (this.props.totalItemsCount - 1) === Number(activeItemIdx) &&
+        Number(activeItemIdx) === this.state.currentIdx;
+
+      switch (code) {
+        case 40: // down
+          if (shouldFocusOutOfViewComponent) {
```
nirnaor: You can remove the long `switch` statement to a one liner 

```
if ((code == 27) || (code === 40 && shouldFocusOutOfViewComponent)) { 

utils.focusGalleryElement(this.props.galleryContainerRef) }
```

packages/lib/src/common/utils/index.js
```diff
@@ -662,6 +662,12 @@ class Utils {
         proGalleryRole === 'application' ? 'gallery application' : 'region',
     };
   }
+
+  focusGalleryElement(element){
+    if(element) {
```
nirnaor: Why do you need the `if`. On which scenario this function is called without an `element`?

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -458,22 +458,37 @@ class SlideshowView extends GalleryComponent {
 
   handleSlideshowKeyPress(e) {
     e.stopPropagation();
-    const nextKeys = [32, 33, 34, 37, 38, 39, 40]
+    const relevantKeys = [32, 33, 34, 37, 38, 39, 40, 27];
     // key code -> 32=space, 37=left, 38=up, 39=right, 40=down
```
ritabazak: add to the comment what "27" is

packages/gallery/src/components/gallery/proGallery/slideshowView.js
ritabazak: it's "utils.focusGalleryElement(this.props.**outOfViewComponent**)" on one condition and "utils.focusGalleryElement(this.props.**galleryContainerRef**)" in the other

packages/gallery/src/components/gallery/proGallery/galleryView.js
yhattav: Hey, please add a comment that explains this rule after we now understand what it means

ritabazak (approved):

PR #351, opened 2021-06-15
add dimensions to item scheme

guysopher:
This will allow the LayoutFixer to create accurate layouts with "fit" crop method

### DIFF
--- a/packages/layouts/src/logic/item.js
+++ b/packages/layouts/src/logic/item.js
@@ -330,6 +330,44 @@ export class Item {
     this.imageMargin = m;
   }
 
+  get dimensions() {
+    //image dimensions are for images in grid fit - placing the image with positive margins to show it within the square
+    const isLandscape = this.ratio >= this.cubeRatio; //relative to container size
+    const imageMarginLeft = Math.round(
+      (this.height * this.ratio - this.width) / -2
+    );
+    const imageMarginTop = Math.round(
+      (this.width / this.ratio - this.height) / -2
+    );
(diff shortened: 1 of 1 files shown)


packages/layouts/src/logic/item.js
```diff
@@ -330,6 +330,44 @@ export class Item {
...
+    } else if (isGridFit && !isLandscape) {
+      dimensions = {
+        //portrait
+        width: this.width - 2 * imageMarginLeft,
+        height: this.height,
+        marginLeft: imageMarginLeft,
+        marginTop: 0,
+      };
+    }
+    return dimensions;
```
nirnaor: Instead of doing a `let` and then mutating the object, you can just move this logic to a function that returns the `dimensions`. 
We try (**nowadays**) to avoid mutations in the code if it isn't necessary since mutations makes debugging harder. 


The (stateless, without `this`) function I'm talking about can get the `isGridFit height width imageMarginLeft  etc` as parameter and simply `return` the dimensions

nirnaor (changes requested):

packages/layouts/src/logic/item.js
nirnaor: @guysopher I wasn't sure if my comment was clear so I pushed a suggestion commit. You can revert it and do it your own way if you don't like it.

guysopher: Looks right.
Made a minor change and pushed

nirnaor (approved):

PR #356, opened 2021-06-24
[gallery] remove video poster and background when needed

IzaacAyelin:
**what** -  removing the posters/background for videos in certain cases (single horizontal item view + auto play) and a small refactor.
**why** - we don't want to display the placeholder/posters in videos in the case mentioned (placeholder was already removed in earlier PR, had to do the same for the video poster and background)

### DIFF
--- a/packages/gallery/src/components/item/videos/videoItem.js
+++ b/packages/gallery/src/components/item/videos/videoItem.js
@@ -315,7 +315,10 @@ class VideoItem extends GalleryComponent {
 
     return (
       <div key={'video-and-hover-container' + this.props.idx}>
-        {[video, videoPlaceholder, hover]}
+        {video}
+        {shouldCreateVideoPlaceholder(this.props.styleParams) &&
+          videoPlaceholder}
+        {hover}
       </div>
     );
   }
--- a/packages/gallery/src/components/item/videos/videoItem.js
+++ b/packages/gallery/src/components/item/videos/videoItem.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { GALLERY_CONSTS, window, utils } from 'pro-gallery-lib';
 import { GalleryComponent } from '../../galleryComponent';
+import { shouldCreateVideoPlaceholder } from '../itemHelper';
 
 class VideoItem extends GalleryComponent {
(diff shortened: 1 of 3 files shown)


packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -231,10 +232,12 @@ class VideoItem extends GalleryComponent {
...
-              poster: this.props.createUrl(
-                GALLERY_CONSTS.urlSizes.SCALED,
-                GALLERY_CONSTS.urlTypes.HIGH_RES
-              ),
+              poster: shouldCreateVideoPlaceholder(this.props.styleParams)
+                ? this.props.createUrl(
+                    GALLERY_CONSTS.urlSizes.SCALED,
+                    GALLERY_CONSTS.urlTypes.HIGH_RES
+                  )
+                : '',
```
nirnaor: I see this pattern coming up a lot in the code: 

1. There will always be a `poster` attribute 
2. When the `poster` isn't relevant it will have a "no meaning" value such as `''`. 


My suggestion is to create a `PosterVideoItem` that extends `VideoItem` and will always have relevant value. 
The `PosterVideoItem` will be created in an upper level based on `shouldCreateVideoPlaceholder`

packages/gallery/src/components/item/videos/videoItemWrapper.js
```diff
@@ -64,11 +65,8 @@ class VideoItemWrapper extends ImageItem {
       'settings',
       'actions',
     ]);
-    const shouldCreatePlaceHolder =
-      utils.isSingleItemHorizontalDisplay(styleParams) && 
-      styleParams.videoPlay === GALLERY_CONSTS.videoPlay.AUTO;
 
-    return shouldCreatePlaceHolder ? null : (
+    return !shouldCreateVideoPlaceholder(styleParams) ? null : (
```
nirnaor: Same here. I think I was supposed to dig in a bit more into this and give you an alternative approach to implement this. Maybe we should dig in to it together.

nirnaor (approved):
You can merge it but I think we should find a solution together to avoid this pattern

PR #363, opened 2021-06-30
Change the behavior of what happens at the first click when the item …

yakovhaiilo:
When the Info Position is set to "on image" and the ״what happens on the first tap״ is set to "no change"  the first tap should be the on-click behavior the user chose in the settings tab.


nirnaor (approved):

packages/gallery/src/components/item/itemView.js
```diff
@@ -198,27 +198,35 @@ class ItemView extends GalleryComponent {
       e.preventDefault();
     }
 
-    if (this.shouldShowHoverOnMobile()) {
+    if (
+      this.shouldShowHoverOnMobile() &&
+      this.props.styleParams.hoveringBehaviour !==
```
IzaacAyelin: maybe use destructure here: const { hoveringBehaviour } = this.props.styleParams

packages/gallery/src/components/item/itemView.js
```diff
@@ -198,27 +198,35 @@ class ItemView extends GalleryComponent {
...
+    const { directLink, styleParams } = this.props;
     const { url, target } = directLink || {};
     const useDirectLink = !!(
       url &&
       target &&
       this.props.styleParams.itemClick === 'link'
     );
-    const shouldUseDirectLinkMobileSecondClick =
+    
+    const constantHoverBehavior = styleParams.hoveringBehaviour ===
```
IzaacAyelin: same destructure here

PR #364, opened 2021-06-30
ArrowsPadding - fix container size to be same as svg size

ritabazak:


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -880,23 +880,26 @@ class SlideshowView extends GalleryComponent {
     }
 
     const { customNavArrowsRenderer } = this.props;
-    let arrowRenderer, navArrowsContainerSize;
+    let arrowRenderer, navArrowsContainerWidth, navArrowsContainerHeight;
```
nirnaor: I prefer to avoid mutations in the code. It makes it much more difficult to track what is changing and by whom. You can do something llike this:
```
{navArrowsContainerWidth, navArrowsContainerHeight} = getHeightAndWidth(customNavArrowsRenderer)
```
The `getHeightAndWidth` function is a pure function that can be outside of the class (can still be in the same file if you want)

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -842,15 +842,68 @@ class SlideshowView extends GalleryComponent {
...
+
+    const arrowOrigWidth = 23; //arrow-right svg and arrow-left svg width
+    const arrowOrigHeight = 39; //arrow-right svg and arrow-left svg height
+    const scalePercentage = arrowsSize / arrowOrigWidth;
+    const svgStyle = { transform: `scale(${scalePercentage})` };
+
+    const navArrowsContainerWidth = arrowsSize; // === arrowOrigWidth * scalePercentage
+    const navArrowsContainerHeight = arrowOrigHeight * scalePercentage;
+
+    const svgInternalStyle = {};
```
nirnaor: Altough javascript allows a mutation of a constant object, I prefer to avoid it as well (since it's a pitfall of the language IMO). you can do something like this instead:
```
const svgInternalStyle = utils.isMobile() && this.?props?arrowsColor? :  {fill: arrowsColor.value} : {}
```
This way you won't mutate the `svgInternalStyle` (even tough JS allows it)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -842,15 +842,68 @@ class SlideshowView extends GalleryComponent {
...
+      const { arrowsColor } = this.props;
+      if (typeof arrowsColor !== 'undefined') {
+        svgInternalStyle.fill = arrowsColor.value;
+      }
+    }
+
+    const arrowRenderer = (position) => {
+      if (position === 'left') {
+        return (
+          <svg width="23" height="39" viewBox="0 0 23 39" style={svgStyle}>
```
nirnaor: I think it's better to reuse the variables you defined earlier:
```
          <svg width={arrowOrigWidth} height={arrowOrigHeight} viewBox={`0 0 {arrowOrigWidth} {arrowOrigHeight`} style={svgStyle}>
```

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -842,15 +842,68 @@ class SlideshowView extends GalleryComponent {
...
+              className="slideshow-arrow"
+              style={svgInternalStyle}
+              d="M154.994,259.522L153.477,261l-18.471-18,18.473-18,1.519,1.48L138.044,243Z"
+              transform="translate(-133 -225)"
+            />
+          </svg>
+        );
+      } else if (position === 'right') {
+        return (
+          <svg width="23" height="39" viewBox="0 0 23 39" style={svgStyle}>
```
nirnaor: The entire SVG is duplicated code :( 
The only difference I see is the `d` and `transform`. You can do something like this instead: 
```
const {d, transform} = position === 'right' ? {d: "M857.005,231.479L858.5,230l18.124,18-18.127,18-1.49-1.48L873.638,248Z",  transform: "translate(-855 -230)" } : {d:"M154.994,259.522L153.477,261l-18.471-18,18.473-18,1.519,1.48L138.044,243Z", transform:"translate(-133 -225)"}
```
And then 
```
return (
          </svg>
            <path
              className="slideshow-arrow"
              style={svgInternalStyle}
              d={d}
              transform={transform}
            />
          </svg>
```

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
ritabazak: Cool, will do. it was just an old code I moved here.

packages/gallery/src/components/gallery/proGallery/slideshowView.js
ritabazak: Yes I see, but the reason those variables were defined as they were (23, 39) is because the SVG itself, as the SVG is sort of a "hardcoded" UI component.

nirnaor:
Ok, but we can make it better :)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
nirnaor: Ok, but we can make it less hardcoded with the constant defined in the same function. It will clean it up a bit :)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -842,15 +842,61 @@ class SlideshowView extends GalleryComponent {
...
+
+    const arrowOrigWidth = 23; //arrow-right svg and arrow-left svg width
+    const arrowOrigHeight = 39; //arrow-right svg and arrow-left svg height
+    const scalePercentage = arrowsSize / arrowOrigWidth;
+    const svgStyle = { transform: `scale(${scalePercentage})` };
+
+    const navArrowsContainerWidth = arrowsSize; // === arrowOrigWidth * scalePercentage
+    const navArrowsContainerHeight = arrowOrigHeight * scalePercentage;
+
+    const { arrowsColor } = this.props;
```
nirnaor, line 864: Since `arrowsColor` isn't used elsewhere you can move use the `this.props.arrowsColor` to next line  if you want :)

nirnaor (approved):

PR #365, opened 2021-07-01
V4 runtime errors from types

nirnaor:
### Why? 
Adding TypeScript to the gallery doesn't protect us from JS projects that are using the gallery. They don't have compilation of TS therefor won't know which API they need to change. 

### What? 

1. Generating schema in build time (input is `d.ts` output is `shcema.js`)
2. Validating that schema in runtime (input is `styleParams` + `schema.js`, output is a list of errors
3. Displaying a list of errors instead of the gallery until no errors exist.

##### Notes

1. Validation needs to be enabled using `shouldValidateTypes` prop or query param.
2. Validation module is dynamically imported, therefor doesn't effect the bundle size.
3. Validation happens on `componentDidMount` since the import has to be dynamic (`async`)
4. [AJV](https://ajv.js.org/) dependency module isn't a part of the bundle since it's part of the dynamically loaded module.


ritabazak (approved):

packages/ssr-simulator/config/webpackConfigFactory.js
```diff
@@ -139,22 +139,22 @@ module.exports = function(envType) {
...
-          }),
-          new OptimizeCSSAssetsPlugin({})
-        ]
-      };
+  // config.optimization = IS_DEV
+  //   ? {}
+  //   : {
+  //       minimizer: [
+  //         new UglifyJsPlugin({
+  //           parallel: true,
```
nirnaor, line 147: @guysopher Turns out `uglifyJS` doesn't support ES6. Since it's an optimisation for a playground project I disabled it's usage. FYI

packages/gallery/src/components/gallery/typeValidator/shouldValidate.js
```diff
@@ -0,0 +1,21 @@
+export default function shouldValidate(props, isSSR) {
```
yhattav, line 1: you can just have 'shouldValidateTypes' (or reinforce the name) as an argument to the function and not have all the props passed to it.

nirnaor: Added to the [next refactoring task ](https://app.asana.com/0/1200247106759123/1200559598586767)

PR #368, opened 2021-07-05
Assuring no style is changed before refactoring

nirnaor:
**Why?**
We have a lot of duplicated style params inside the WIX repository. Doing a breaking change in these params will obligate us to change all across the wix-private repository.
Instead, if we will reuse the duplicated styles, the only place we will need to refactor is the core repository (and WIX will get it automatically)

**What?**
Adding a dummy test that assures that the JSON remains the same 
Moving the duplicated styles to a separate file and exporting it for wix-private usage.

This is a prerequisite for another PR that will be inside the wix private repo


packages/lib/test/defaultStyles.spec.js
```diff
@@ -0,0 +1,107 @@
+import { expect } from 'chai';
+import _ from 'lodash';
```
ritabazak, line 2: Do not import the whole lodash package (also remove from package.json).
Implement the needed functions here: 
https://github.com/wix/pro-gallery/blob/master/packages/lib/src/common/utils/lodash.js
and import it.

packages/lib/src/index.ts
```diff
@@ -3,6 +3,7 @@ export { default as GalleryItem } from './core/galleryItem';
 export { default as BlueprintsManager } from './core/blueprints/BlueprintsManager';
 export { default as blueprints } from './core/blueprints/Blueprints';
 export { default as defaultStyles } from './common/defaultStyles';
+export { default as duplicatedInWix } from './common/duplicatedInWix';
```
ritabazak: Maybe should be renamed to "duplicatedStylesInWix" or something like that, cause when imported from "pro-gallery" it will not be cleated what is the "thing" that is duplicated.

ritabazak: We also try to not use the word "wix" in the core, don't have a better idea for naming but it's something to think about.

packages/lib/test/defaultStyles.spec.js
nirnaor: It's a dev dependency therefor doesn't effect bundle size.

packages/lib/src/index.ts
nirnaor: I can call it coreStyles instead of using wix.

PR #369, opened 2021-07-06
5 july missing navigation arrows

yakovhaiilo:


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -78,7 +78,7 @@ class SlideshowView extends GalleryComponent {
       return false;
     }
 
-    if (this.container) {
+    if (this.scrollElement) {
```
nirnaor: This condition is duplicated inside `this.getScrollLeft()`, therefor the same "work" (checking `this.scrollElement` is being done twice :(

nirnaor: Since `this.getScroolLeft` returns 0 when `this.scrollElement` is `undefined` you can use that function to `return false` (instead of duplicating the tests:
```
isScrollStart() {
    return this.props.styleParams === GALLERY_CONSTS.slideAnimations.SCROLL) || this.getScrollLeft() > 1
}
```

nirnaor (changes requested):

nirnaor (approved):

PR #373, opened 2021-07-07
[gallery] fix: use imageDimensions in video item

IzaacAyelin:
Fix videos and text items positioning and dimensions (following the image dimensions refactor to the layouter)


packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -280,13 +279,16 @@ class VideoItem extends GalleryComponent {
           utils.deviceHasMemoryIssues() ||
           this.state.ready ||
           !shouldCreateVideoPlaceholder(this.props.styleParams)
-            ? { backgroundColor: 'black' }
+            ? {
```
nirnaor: There's too much code in the JSX, I prefer to avoid ocmplicated JSX. There are two things I would do here:

1. Reuse the duplicated `imageDimensions` (make it your start value that you change based on the condition
2. Move the entire style calculation to a function (not inside the JSX) 
3. If `this.state.ready` || `utils.deviceHasMemoryIssues()` - we don't want any styles at all? Is that the case?

packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -269,8 +269,7 @@ class VideoItem extends GalleryComponent {
       baseClassName += ' playedOnce';
     }
     // eslint-disable-next-line no-unused-vars
-    const { marginLeft, marginTop, ...restOfDimensions } =
-      this.props.imageDimensions || {};
+    const imageDimensions = this.props.imageDimensions || {};
```
nirnaor: In which case we don't have the `imageDimensions` prop?

nirnaor (changes requested):

packages/gallery/src/components/item/videos/videoItem.js
IzaacAyelin: 1,2 - OK, sounds better
3 - in that case we don't want the background image

packages/gallery/src/components/item/videos/videoItem.js
IzaacAyelin: I don't think we have a case like that, so I guess this empty object Is useless

IzaacAyelin: it seems that i was wrong, we don't always have it at first

IzaacAyelin: at least with the unit tests (maybe they are not written well)

PR #385, opened 2021-07-15
Ignore first scroll event if the gallery opens in a specific idx

guysopher:


guysopher:
This branch is used to use the blurry image or placeholder if the fullscreen is opened with a specific id

nirnaor (approved):

ritabazak:
change "currentIdx" to "activeIndex"

guysopher:
The change will make the fullscreen load with the preloaded blurry image

Since the fullscreen is now using the gallery, when you open the fullscreen, it is using the load mechanism that prevents the blurry images from loading until the first scroll event. So when you open the fullscreen, the first image takes a long time to load even though the blurry image is already loaded in the browser.
The change will assume that if you passed a currentIdx, it is equal to a scroll event, and the preloaded image will be shown

guysopher:
> change "currentIdx" to "activeIndex"

Done

PR #396, opened 2021-07-21
V4: rename "domId" to "id"

ritabazak:


ritabazak:
renamed `domId` to `id`, changed `totalItemsCount` to be not mandatory.

nirnaor (approved):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -106,14 +106,14 @@ export class GalleryContainer extends React.Component {
       );
     }
 
-    if (this.props.domId) {
-      this.currentHoverChangeEvent.domId = this.props.domId;
+    if (this.props.id) {
+      this.currentHoverChangeEvent.galleryId = this.props.id;
```
yhattav, line 110: This is internal to that gallery?

ritabazak: Yes, cause itemView already had "id" as a prop (the actual item id), so to itemView I pass the "id" of the gallery as "galleryId".
itemView is the part of the code that is actually using this event and the event data.

PR #421, opened 2021-07-28
Sp refactor preset function

yhattav:


packages/lib/src/core/helpers/layoutHelper.js
```diff
@@ -35,6 +35,59 @@ export const calcTargetItemSize = (styles, smartCalc = false) => {
   }
 };
 
+export const mutateNumberOfImagesPerRow = (styles) => {
+  //This will be used in the masonry and grid presets
+  let res = {...styles}
+  if (
+    styles.scrollDirection === GALLERY_CONSTS.scrollDirection.VERTICAL || //relevant for grid, in Masonry its fixed to !oneRow
```
nirnaor, line 42: I prefer to avoid mutation when possible
```
const res = {...styles}
if (styles.scrollDirection === GALLERY_CONSTS.scrollDirection.VERTICAL) {
    return Object.assign({}, res, ...)
)
return res;

File: packages/lib/src/core/helpers/layoutHelper.js [hunk 2]
```diff
@@ -35,6 +35,59 @@ export const calcTargetItemSize = (styles, smartCalc = false) => {
...
+    res.groupSize = 1;
+    res.collageAmount = 0;
+    res.collageDensity = 0;
+  } 
+  return res;
+}
+
+export const mutateNumberOfImagesPerCol = (styles) => {
+  //This will be used in the grid preset
+  let res = {...styles}
```
nirnaor, line 59: it actually doesn't mutate if I'm not mistaken, but creates a new object

packages/lib/src/core/presets/magicGallery.js
```diff
@@ -2,35 +2,41 @@ import LAYOUTS from '../../common/constants/layout';
...
-  isSlider: undefined,
-  isColumns: undefined,
-  cropOnlyFill: undefined,
-  fixedColumns: undefined,
-  enableInfiniteScroll: undefined,
-  slideshowLoop: false,
+const fixToMagic = (styles) => {
+  let presetStyles = { ...styles };
+  presetStyles.galleryLayout = LAYOUTS.MAGIC;
+  presetStyles.cubeImages = undefined;
```
nirnaor, line 8: I prefer:
`['cubeImages', 'cubeRatio', ...].forEach(p => delete presetStyles[p])`

nirnaor (approved):

packages/lib/src/core/presets/magicGallery.js
yhattav: All I can do is change this to delete commands. (and I dont think its worth the change for now)
this about the next step where Ill need to delete / assign undefined to a.b.c.d.cubeSomething (as a deeper key)

nirnaor: It's just a style suggestions to make it shorter. The PR is great :)

PR #458, opened 2021-08-04
Fixed cyclic dependency: pro-layouts -> pro-gallery-lib -> pro-layouts

nirnaor:
Why?
A lot of our build scripts can be replaced with Lerna's conventions. 

What?

1. Fix all lerna's hoisting warnings
2. Run lerna in strict mode - warnings now fail the build
3. Deprecating `build:dev build-packages-by-order link-packages`


ritabazak:
It's very cool that you let lerna do its own thing.
Not sure what in this PR is fixing the cyclic dependency, can you please explain? Thanks :)

nirnaor:
@ritabazak That's an old commit message. I already fixed the cycle in a [previous commit](https://github.com/wix/pro-gallery/commit/9ec409c8d9cfadcf1a2ce332a987c2a91b9f3535)

ritabazak:
Ohhh It is the PR name so I got confused..
So happy you removed this cyclic dep!

PR #484, opened 2021-08-09
Navigation arrows vertical positioning

yoavva:


nirnaor (approved):
In the future please add a short description of the code changes.

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -985,14 +965,47 @@ class SlideshowView extends GalleryComponent {
...
       scrollDirection === GALLERY_CONSTS.scrollDirection.HORIZONTAL &&
       arrowsPosition === GALLERY_CONSTS.arrowsPosition.OUTSIDE_GALLERY
         ? `-${20 + navArrowsContainerWidth}px`
         : `${imageMargin / 2 + (arrowsPadding ? arrowsPadding : 0)}px`;
+
+    // arrows on info? True : False
+    const arrowsOnInfo = arrowsVerticalPosition === GALLERY_CONSTS.arrowsVerticalPosition.INFO_CENTER;
+
+    // arrows too big for info? True : False
+    const arrowsDontFitInfo = isSlideshow ?
```
yhattav: use infoHeight

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -985,14 +965,47 @@ class SlideshowView extends GalleryComponent {
...
+    arrowsOnInfo && (slideshowInfoSize < navArrowsContainerHeight) :
+    GALLERY_CONSTS.isAboveOrBelowPlacement(titlePlacement, isSlideshow) && arrowsOnInfo && (textBoxHeight < navArrowsContainerHeight);
+
+    // Should not render nav arrows? True : False
+    const shouldNotRenderNavArrows =
```
yhattav: create a function (without this)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -985,14 +965,47 @@ class SlideshowView extends GalleryComponent {
...
+          0
+        );
+        const isAllItemsFitsGalleryWidth =
+          scrollDirection === GALLERY_CONSTS.scrollDirection.HORIZONTAL &&
+          this.props.container.galleryWidth >= allGroupsWidth;
+        return isAllItemsFitsGalleryWidth;
+      });
+      
+      //remove navBars if no scroll is needed and is column layout
+      if (shouldNotRenderNavArrows) {
```
yhattav: place back in the top of the function

yhattav (changes requested):

packages/gallery/tests/e2e/styleParams/arrowsVerticalPosition.e2e.spec.js
```diff
@@ -0,0 +1,78 @@
...
+  });
+  it('should have arrows on image center and info above', async () => {
+    await driver.navigate({
+      galleryLayout: GALLERY_CONSTS.layout.SLIDER,
+      titlePlacement: 'SHOW_ABOVE',
+      arrowsVerticalPosition: 'IMAGE_CENTER',
+    });
+    await driver.waitFor.hookToBeVisible('item-container');
+    const page = await driver.grab.elemScreenshot('.pro-gallery');
+    expect(page).toMatchImageSnapshot();
```
nirnaor: These 3 lines are duplicated across all tests. Please reuse

packages/gallery/tests/e2e/styleParams/arrowsVerticalPosition.e2e.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      titlePlacement: 'SHOW_ABOVE',
+      arrowsVerticalPosition: 'INFO_CENTER',
+    });
+    await driver.waitFor.hookToBeVisible('item-container');
+    const page = await driver.grab.elemScreenshot('.pro-gallery');
+    expect(page).toMatchImageSnapshot();
+  });
+  it('should have arrows on item center and info above', async () => {
+    await driver.navigate({
+      galleryLayout: GALLERY_CONSTS.layout.SLIDER,
```
nirnaor: This `galleryLayout` is duplicated across all tests. Please reuse

packages/gallery/tests/styleParams/arrowsVerticalPosition.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+  afterEach(() => {
+    driver.detach.proGallery();
+  });
+
+  it('Should not render the arrows when not enough info space', async () => {
+    initialProps.options = mergeNestedObjects(initialProps.options, {
+      galleryLayout: GALLERY_CONSTS.layout.SLIDESHOW,
+      slideshowInfoSize: 38, // info height is 39 when default, so slideshowInfoSize < arrowsSize
+      arrowsVerticalPosition: 'INFO_CENTER',
```
nirnaor: These props are duplicated across all tests. Please reuse.

packages/lib/src/core/helpers/layoutHelper.js
```diff
@@ -279,6 +281,20 @@ const addMarginsToSupportShadows = (options) => {
   }
   return _options;
 }
+const centerArrowsWhenNeeded = (options) => {
+  let _options = {...options}
+  const filteredPlacement = _options.titlePlacement // filtering hover since it doesn't affect this product
+  .split(',')
+  .filter(placement => !isHoverPlacement(placement))
```
nirnaor, line 288: nice

nirnaor (changes requested):

packages/lib/src/core/helpers/layoutHelper.js
yoavva: Thanks!

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -913,6 +932,47 @@ class SlideshowView extends GalleryComponent {
...
+      arrowsVerticalPosition,
+      textBoxHeight,
+    } = this.props.options;
+    const { height } = this.props.container;
+    // Calc of Nav arrows container's height
+    const { navArrowsContainerHeight } = this.getArrowsSizeData();
+    let parentHeight;
+    const infoHeight = isSlideshow ? slideshowInfoSize : textBoxHeight;
+    switch (arrowsVerticalPosition){
+      case GALLERY_CONSTS.arrowsVerticalPosition.INFO_CENTER:
```
nirnaor: Please change this `switch` to a map. Ask @yuvalab about it (she just refactored something similar now)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -925,53 +985,42 @@ class SlideshowView extends GalleryComponent {
...
+    const galleryVerticalCenter = `50% + ${imageMargin/4}px` 
+
+    // Determines the direction fix, the direction in which we move the nav arrows 'vertical position fix' pixels 
+    let directionFix;
+    if (GALLERY_CONSTS.hasExternalAbovePlacement(titlePlacement)){
+      directionFix = -1;
+    } else if (isSlideshow || GALLERY_CONSTS.hasExternalBelowPlacement(titlePlacement)){
+      directionFix = 1;
+    } else {
+      // if we got here, we should be ITEM_CENTER, taken care of in layoutHelper.js
```
nirnaor, line 908: ????

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -913,6 +932,47 @@ class SlideshowView extends GalleryComponent {
...
+        parentHeight = height;
+        break;
+    }
+    return parentHeight >= navArrowsContainerHeight;
+
+  }
+
+  // function to Determine whether we should render the navigation arrows
+  shouldRenderNavArrows = () => {
+    const { showArrows } = this.props.options;
```
nirnaor: Please move all the navigation arrows methods you added to a new file. Call it `navigationArrowUtils`. 
The functions should have no dependency in `this`.
Pass each function the relevant arguments it needs.
I want the navigation arrows logic separated from the slideshowView file.  (since HOC refactoring is too much for now, then a simple separation will be enough at the moment)

packages/gallery/src/components/gallery/proGallery/slideshowView.js
nirnaor: throw error or use default if you don't know what to do

packages/gallery/tests/styleParams/arrowsVerticalPosition.spec.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('Checks if "IMAGE_CENTER" has correct distance from top for SlideShow', async () => {
+    // Exact style params relevant to this sub-test
+    initialProps.options = mergeNestedObjects(initialProps.options, {
+      arrowsVerticalPosition: 'IMAGE_CENTER',
+    });
+    driver.mount.proGallery(initialProps);
+    await driver.update();
+    const navArrows = driver.find.selector('.nav-arrows-container');
```
nirnaor: These lines are duplicated across 3 tests :(

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -863,54 +864,34 @@ class SlideshowView extends GalleryComponent {
...
-    const { customNavArrowsRenderer } = this.props.customComponents;
-    const { arrowsSize } = this.props.options;
-    if (customNavArrowsRenderer) {
-      return {
-        arrowRenderer: customNavArrowsRenderer,
-        navArrowsContainerWidth: arrowsSize,
-        navArrowsContainerHeight: arrowsSize
-      }
+  
+  getShouldRenderArrowsArgs(){
```
nirnaor: `getShouldRenderArrowsArgs` is arrows related, please move the the `navigationArrowUtils` file

packages/gallery/tests/e2e/styleParams/arrowsVerticalPosition.e2e.spec.js
```diff
@@ -0,0 +1,101 @@
...
+    expect(page).toMatchImageSnapshot();
+  });
+
+  afterAll(async () => {
+    await driver.closePage();
+  });
+
+  it('should have arrows on info center', async () => {
+    const currentTestLayout = {
+      ...layoutStyles,
```
nirnaor: The only difference between the tests is the `arrowsVerticalPosition`. Don't add the `...layoutStyles` again on every test. Reuse it please

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1147,7 +1096,7 @@ class SlideshowView extends GalleryComponent {
         style={galleryStyle}
       >
         {this.createDebugMsg()}
-        {this.createNavArrows()}
+        {shouldRenderNavArrows(getShouldRenderArrowsArgs(this.props)) && this.createNavArrows()}
```
nirnaor: `getShouldRenderArrowsArgs` can be turned to a private method of the `navigationArrowUtils` you created. When a module (or even file)  such as `navigationArrowUtils` contains fewer exports, it makes it easier to understand the dependencies of the module. Private methods are awesome (private === not exported).
So, here's what you can do: 

1. Pass `container` and `options` to `shouldRenderNavArrows`
2. Pass `options` to `getShouldRenderArrowsArgs`
3. Remove the `export` of `getShouldRenderArrowsArgs` - since it's not used elsewhere
4. Celebrate

nirnaor (changes requested):

packages/gallery/tests/e2e/styleParams/arrowsVerticalPosition.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+    await driver.navigate(currentTestLayout);
+  });
+
+  it('should have arrows on image center', async () => {
+    const currentTestLayout = getLayoutStylesByVerticalPosition(
+      'SHOW_BELOW',
+      'IMAGE_CENTER'
+    );
+    await driver.navigate(currentTestLayout);
```
nirnaor, line 69: @yoavva are you sure something is being tested here?

packages/gallery/src/components/gallery/proGallery/slideshowView.js
yoavva: Great explanation, thanks!
I've turned getShouldRenderArrowsArgs to private method, mind taking a look?
Please note that shouldRenderNavArrows must take a few more arguments (other than options and container), therefore some of the deconstruction happens now in slideshowView. I think it's a tradeoff (turning getShouldRenderArrowsArgs private <-> amount of deconstructing in slideshowView) I can't avoid, do you have an idea on how to avoid it? does the current change good enough?
thanks again for the explanation!
@nirnaor

nirnaor: As a rule of thumb I try to pass the minimum amount of dependencies (arguments in your case) from the top level to the bottom (from `sliedeshowView` to `navigationArrowUtils` in your case). Therefor, destructing at the higher level (`slideshowview`) **is better** (like you did). So I think you are good on that front.

packages/gallery/tests/e2e/styleParams/arrowsVerticalPosition.e2e.spec.js
yoavva: My bad! they weren't sensitive enough, they are now
@nirnaor

nirnaor: 1. What do you mean they are now? 
2. What do you mean sensitve?
3.  There's no expect that I'm seeing in this test?
It doesn't test anything @yoavva

yoavva: 3. The expect is in the afterEach methods, I've put it there to prevent code duplication. 
1-2. The tests were not sensitive enough earlier which means that even if the arrows were not located where they should have, this change was small enough so that the tests will pass anyway (toMatchImageSnapshot was configured with 5% difference threshold so the arrows dislocated made less than 5% difference between the images due to their small size I guess) , however, I've added 'testsSensitivityConfig' to toMatchImageSnapshot(testsSensitivityConfig), which makes the tests not tolerable to any change and now they will fail the test for every minor dislocation of the nav arrows.
@nirnaor

nirnaor: @yoavva 
`afterEach` is used to clean **after the test is finished**. You should not use it for assertions. 
I'm not even sure If tests will be marked as failures if their teardown fails. 
You need to `expect` inside the test. Just use a function that does that, but not `afterEach`.
P.S
I'm not a fan of `beforeEach` and `afterEach`. I think it makes the reading experience of tests terrible. I just prefer to reuse functions and to read each test from start to bottom.

nirnaor: More about [it here](https://jestjs.io/docs/setup-teardown)

yoavva: Thanks for the explanation! 
Got it, no more testing in those functions. 
Changed it, can you take a look?
@nirnaor

packages/gallery/tests/styleParams/arrowsVerticalPosition.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      arrowsVerticalPosition: 'INFO_CENTER',
+    });
+    const navArrows = await mountGalleryAndGetArrows(initialProps);
+    const galleryContainer = driver.find.selector('#pro-gallery-container');
+    const { height: galleryHeight } = galleryContainer.props().style;
+    const { top } = navArrows.props().style;
+    const expectedInfoSpace = (-1 * galleryHeight) / 2;
+    expect(top.replace(/\s/g, '')).to.eq(
+      `calc(50%+0px-19.5px-${expectedInfoSpace}px)`
+    );
```
nirnaor: Duplicated to lines 81 - 84

packages/gallery/tests/styleParams/arrowsVerticalPosition.spec.js
```diff
@@ -0,0 +1,86 @@
...
+  it('Checks if "IMAGE_CENTER" has correct distance from top for SlideShow', async () => {
+    // Exact style params relevant to this sub-test
+    initialProps.options = mergeNestedObjects(initialProps.options, {
+      arrowsVerticalPosition: 'IMAGE_CENTER',
+    });
+    const navArrows = await mountGalleryAndGetArrows(initialProps);
+    const { top } = navArrows.props().style;
+    const expectedInfoSpace = initialProps.options.slideshowInfoSize / 2;
+    expect(top.replace(/\s/g, '')).to.eq(
+      `calc(50%+0px-19.5px-${expectedInfoSpace}px)`
```
nirnaor: At the end of this refactoring this complicated `calc` should be written only one time in this file

nirnaor (changes requested):

packages/gallery/tests/styleParams/arrowsVerticalPosition.spec.js
yoavva: Done 
@nirnaor

nirnaor (approved):

PR #521, opened 2021-08-15
Magnified images feature

IzaacAyelin:
A new feature that allows users to magnify images on click


packages/gallery/src/components/item/itemView.js
```diff
@@ -22,6 +22,9 @@ import {
   getImageStyle,
 } from './itemViewStyleProvider';
 import VideoItemWrapper from './videos/videoItemWrapper';
+import withMagnified from '../hoc/withMagnified.js';
+
+const MagnifiedImageItem = withMagnified(ImageItem);
```
nirnaor: @IzaacAyelin 
Why does the `itemView` knows the magnifier? Didn't we say that because `itemView` can be a video or text - the magnifier will be inside of `imageItem`?

nirnaor (changes requested):

packages/gallery/src/components/item/itemView.js
IzaacAyelin: we have some other places that the ImageItem that I didn't notice at first, so I couldn't export it directly with the hoc

nirnaor (approved):

packages/lib/src/common/constants/itemTypes.js
```diff
@@ -0,0 +1,9 @@
+const ITEN_TYPES = {
```
ritabazak: Please change to `ITEM_TYPES`

ritabazak: And if I remember correctly, there is also 'photo', no?

PR #524, opened 2021-08-15
Last lint

nirnaor:
**Why?**

1. Every commit ran REALLY slow because the `eslint --fix` wasn't configured correctly
2. The configuration was not scalable. Every package added (like today's `pro-gallery-blueprints` required me to add another `precommit:blueprints` script and another `lint-staged` configuration

**What?**

1. Remove the old precommits scripts - they are not needed anymore
2. Use a more robust glob pattern - so new packages will be detected automatically without more configuration.

Notes:

- `lint-staged` is passing the [absolute file paths](https://github.com/okonet/lint-staged#filtering-files) eslint command, allowing me to run the command from the root
- `eslint` configuration is [cascading](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy), therefor the configuration for each files is determined by it's location on the file system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   "lint-staged": {
     "packages/layouts/{src,test}/**/*.{js,ts}": "npm run precommit:layouts",
     "packages/lib/{src,test}/**/*.{js,ts}": "npm run precommit:lib",
-    "packages/gallery/{src,tests}/**/*.{js,ts,tsx}": "eslint --fix --",
+    "packages/*/{src,tests}/**/*.{js,ts,tsx}": "eslint --fix --",
     "packages/playground/src/**/*.{js,ts}": "npm run precommit:playground",
     "packages/ssr-simulator/src/**/*.{js,ts}": "npm run precommit:ssr-simulator",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -55,12 +48,7 @@
     }
   },
   "lint-staged": {
-    "packages/layouts/{src,test}/**/*.{js,ts}": "npm run precommit:layouts",
-    "packages/lib/{src,test}/**/*.{js,ts}": "npm run precommit:lib",
-    "packages/gallery/{src,tests}/**/*.{js,ts,tsx}": "npm run precommit:gallery",
-    "packages/playground/src/**/*.{js,ts}": "npm run precommit:playground",
-    "packages/ssr-simulator/src/**/*.{js,ts}": "npm run precommit:ssr-simulator",
-    "packages/lean-gallery/src/**/*.{js,ts}": "npm run precommit:lean-gallery"
+    "packages/*/{src,tests}/**/*.{js,jsx,ts,tsx}": "eslint --fix --"
```
ritabazak, line 51: This config is not covering the "test" folder of layouts and of lib.
@yhattav @nirnaor

nirnaor: good catch @ritabazak

PR #534, opened 2021-08-18
refactor next item and next group

yakovhaiilo:
We wanted to improve the logic in the next functions (next, "nextItem", "nextGroup")

Why?
The scroll continues to play, even if we have reached the last item, and some of the questions asked in the function were incorrect or in the right place. 

What? 
- We check if next is supposed to do an action, based on the scroll direction and the item defined in the state **shouldBlockNext**

- Asked more correct question whether a scroll should be done **shouldNotAllowScroll**

expected : 
That the code will be more readable and understandable in the parts we have improved
That the scrolling will not be done if there is nowhere to scroll
Delete unnecessary code


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -337,35 +341,14 @@ class SlideshowView extends GalleryComponent {
...
-        !isAutoTrigger &&
-        ((direction >= 1 && this.isLastItemFullyVisible()) ||
-          (direction <= -1 && this.isFirstItemFullyVisible()));
       const scrollMarginCorrection = this.getStyles().margin || 0;
       const _scrollDuration =
         scrollDuration || this.props.styleParams.scrollDuration || 400;
+      const shouldAllowScroll = !this.shouldNotAllowScroll({ scrollingUpTheGallery });
 
-      !isScrollingPastEdge &&
+      shouldAllowScroll &&
```
nirnaor, line 351: The entire code in lines 278-329 (**lets call it code A**) is duplicated between the lines and 345-379 (**lets call it code B**). 

Here is the list of  **differences** between these two chunks of code is (**everything else is duplicated**)

1. Code A calls `scrollToItem`. Code B calls `scrollToGroup`. (Note that the arguments passed there are **also duplicated**)
2. There's the `this.props.styleParams.groupSize === 1` handling only in code A
3. The `activeIndex` has different values between code A and code B
4. There's an `ignoreScrollPosition` handling in code A
5. The `console.error` has different text in the `catch` block

nirnaor (changes requested):
Please extract duplicated logic to a method, reuse it in both places. The only difference between `nextItem` and `nextGroup` should be the 5 bullets mentioned in the comment

packages/gallery/src/components/gallery/proGallery/slideshowView.js
yhattav: Hey Yakov & Nir,
This refactor was about the block of next and the block of the scroll itself in two functions, this is why Yakov did not touch other possible refactors.

Yakov - let's finish this up with Nirs list to leave a really clean code.

PR #552, opened 2021-08-23
[gallery] change rendererIndex to be a tsx file

ritabazak:


packages/gallery/src/components/gallery/galleryRendererTypes.ts
```diff
@@ -0,0 +1,7 @@
+export interface GalleryRendererProps {
+  container: any;
```
nirnaor: @ritabazak Please place all files that contains interfaces only (not logic) inside the `common/interfaces` folder

ritabazak: In this PR? or another PR?

nirnaor: I don't see any value in creating a separate PR

yhattav (approved):

PR #575, opened 2021-08-29
moving props.

nirnaor:


nirnaor:
@yhattav Playground video playTrigger is working with this PR. We can review it together when you have time

packages/lib/src/common/constants/videoPlay.js
```diff
@@ -1,7 +1,7 @@
 const VIDEO_PLAY = {
-  HOVER: 'hover',
-  AUTO: 'auto',
-  ON_CLICK: 'onClick',
```
nirnaor, line 4: @yhattav I think it might be related to the failing specs

PR #634, opened 2021-10-06
let error propagate

IzaacAyelin:


IzaacAyelin:
@nirnaor you think it would be better to just let the error propagate or console.error it after try/catch

nirnaor:
@IzaacAyelin What happens when you remove the entire `catch` ?

IzaacAyelin:
@nirnaor nothing will happen

nirnaor (approved):

PR #638, opened 2021-10-10
fix first video playing when not needed

IzaacAyelin:
**What** - initialize videoScrollHelperWrapper with the correct top/left values 
**why** - when activeIndex is not 0 the first video is still playing

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -81,7 +81,7 @@ export class GalleryContainer extends React.Component {
   initializeScrollPosition() {
     if (this.props.activeIndex > 0) {
       this.scrollToItem(this.props.activeIndex, false, true, 0);
-      const currentItem = this.galleryStructure.items[this.props.activeIndex || 0];
+      const currentItem = this.galleryStructure.items[this.props.activeIndex];
       this.onGalleryScroll(currentItem.offset);
     }
   }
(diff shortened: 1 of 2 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -84,6 +84,8 @@ export class GalleryContainer extends React.Component {
     this.handleNewGalleryStructure();
     this.eventsListener(GALLERY_CONSTS.events.APP_LOADED, {});
     this.videoScrollHelper.initializePlayState();
+    const currentItem = this.galleryStructure.items[this.props.activeIndex || 0];
+    this.videoScrollHelper.initScroll(currentItem.offset);
```
yhattav: why do you need all this new code. you have a function for this called onGalleryScroll

yhattav (changes requested):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
IzaacAyelin: it just didn't feel right calling the onGalleryScroll (which also fires an event - GALLERY_SCROLLED) since all I wanted was to initialize the left/top with the right values

yhattav: but you are scrolling the gallery to the active index. you do this in the component did mount. so actually you are scrolling and when you scroll an event should fire. so even further... why is the onGalleryScroll not firing in the scrollTo?

nirnaor (approved):

PR #651, opened 2021-11-02
Videos are not playing in galleries below fold (AUTO play)

IzaacAyelin:
**What** - remove condition for vertical scroll listener in galleryScrollIndicator
**Why** - horizontal galleries must also listen to vertical scroll to determine if the gallery is in view and play videos in auto play.

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
@@ -111,7 +111,11 @@ export default class ScrollIndicator extends React.Component {
         this.setState({
           scrollTop: top,
         });
-        this.props.getMoreItemsIfNeeded(top);
+        if (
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
```diff
@@ -76,48 +75,48 @@ export default class ScrollIndicator extends React.Component {
...
+          .horizontal()
+          .addEventListener('scroll', this.onHorizontalScroll);
+
+        scrollingElement
+          .horizontal()
+          .addEventListener(
+            'scrollTransition',
+            this.onHorizontalScrollTransition
+          );
+      } catch (e) {
```
yhattav, line 99: no no no

yhattav (changes requested):

yhattav (approved):

PR #662, opened 2021-11-17
handle text item - accessibility

yuvalab:


packages/gallery/src/components/item/itemHelper.js
```diff
@@ -75,3 +75,10 @@ export function shouldCreateVideoPlaceholder(options) {
       options.videoPlay !== GALLERY_CONSTS.videoPlay.AUTO)
   );
 }
+
+export function extractContent(html) {
```
IzaacAyelin: maybe we should call it `extractTextItemContent`

packages/gallery/src/components/item/itemView.js
```diff
@@ -834,7 +835,8 @@ class ItemView extends React.Component {
         label = '';
         break;
       case 'text':
-        label = 'Text item';
+        label =  extractContent(this.props.html) || 'Text item';
```
IzaacAyelin: I think we can move `|| 'Text item'` to the inside of function
`return content.trim() || 'Text item'`

PR #664, opened 2021-11-18
[gallery] improve: screen readers  can read text item

yuvalab:
accessibility bug- in text item they couldn't hear the right content, but only 'text item' . extract the content allow them to hear the right content.


packages/gallery/src/components/item/itemHelper.js
```diff
@@ -75,3 +75,10 @@ export function shouldCreateVideoPlaceholder(options) {
       options.videoPlay !== GALLERY_CONSTS.videoPlay.AUTO)
   );
 }
+
+export function extractTextItemContent(html) {
+  const span = document.createElement('span');
+  span.innerHTML = html;
+  const content = span.innerText;
+  return content?.trim() || 'Text item';
```
nirnaor: @yuvalab at which case there's no content ?

yuvalab: i thought in case the content is undefined, but i think there is no case like this. so i can delete.

packages/gallery/src/components/item/itemView.js
```diff
@@ -789,7 +789,7 @@ class ItemView extends React.Component {
         label = '';
         break;
       case 'text':
-        label = 'Text item';
+        label = extractTextItemContent(this.props.html);
```
nirnaor: @yuvalab 
Can you change the entire switch case to an object? Same as we did before?

nirnaor (changes requested):

nirnaor (approved):
😍

PR #665, opened 2021-11-24
 accessibility- unique gallery id 

yuvalab:
Using the same ID on more than one element may cause screen readers and other assistive technologies to only announce the first element with the shared ID, preventing users from accessing the later elements.


packages/gallery/tests/drivers/reactDriver.js
```diff
@@ -178,6 +178,10 @@ class galleryDriver {
     };
   }
 
+  getContainer() {
+    return this.find.selector('#pro-gallery-container-default-dom-id');
```
nirnaor: @yuvalab 
What will happen here when the someone in the test will define an ID as a prop ?

nirnaor (changes requested):

packages/gallery/tests/drivers/reactDriver.js
yuvalab: you mean i need to let props.id  || the default string?

nirnaor: Yes

nirnaor: If the user decides in his test that he wants to use an `id` in props - the `getContainer` breaks

nirnaor (approved):

PR #679, opened 2021-12-13
Youtube not playing on first tap on ios

shmuelhizmi:


packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -22,10 +30,19 @@ class VideoItem extends React.Component {
     };
   }
 
+  playingFromIframeClick = false;
```
nirnaor: Why do you hold a property inside a component. It should be dependent on `state` and `props` only

packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -81,13 +97,60 @@ class VideoItem extends React.Component {
     return this.isHLSVideo() && utils.isiOS();
   }
 
+  updatePlayer = async () => {
+    if (!this.isYoutubeVideo()) {
+      return;
+    }
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
nirnaor: This functionality should be injected using hooks or [HOC](https://github.com/wix-private/photography-pro-gallery/pull/909#discussion_r767546350). The word `youtube` should not be written here.

packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -22,10 +30,19 @@ class VideoItem extends React.Component {
...
 
+  playingFromIframeClick = false;
+  pausingFromIframeClick = false;
+
   componentDidMount() {
     this.dynamiclyImportVideoPlayers();
   }
 
+  isVimeoVideo = () =>
+    this.props.videoUrl && this.props.videoUrl.includes('vimeo.com');
```
nirnaor: This code is duplicated. You can do it better

packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -81,13 +97,60 @@ class VideoItem extends React.Component {
...
+      if (!this.state.shouldPlay && this.pausingFromIframeClick) {
+        this.pausingFromIframeClick = false;
+      }
+      if (!this.state.shouldPlay && this.playingFromIframeClick) {
+        break;
+      }
+      if (this.state.shouldPlay && this.pausingFromIframeClick) {
+        break;
+      }
+      if (this.video?.player?.player?.player?.playVideo) {
```
nirnaor: All these conditions won't be needed in case this code runs for youtube only. Correct?

packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -81,13 +97,60 @@ class VideoItem extends React.Component {
...
+
   UNSAFE_componentWillReceiveProps(nextProps) {
     if (nextProps.shouldPlay || nextProps.firstUserInteractionExecuted) {
       this.setState({ loadVideo: true });
     }
 
-    if (nextProps.shouldPlay) {
-      this.setState({ shouldPlay: true });
+    if (nextProps.shouldPlay && !this.state.shouldPlay) {
+      this.playingFromIframeClick = false;
```
nirnaor: You shouldn't alter `this`. Use only `props` and state for the behaviour of your component.

nirnaor (changes requested):

packages/gallery/src/components/item/videos/videoItem.js
shmuelhizmi: the reason I chose to store this variable outside the state is that it does not effect rendering, and only effect the logic of the youtube controller which is independent of the react rendering process

packages/gallery/src/components/item/videos/videoItem.js
shmuelhizmi: I will move must of this code to a separate helper function

packages/gallery/src/components/item/videos/videoItem.js
shmuelhizmi: sure

PR #687, opened 2021-12-23
settings  icons and test

IzaacAyelin:
Added new option - arrows type (layoutParams_navigationArrows_type) + new icons
New navigation arrows options for horizontal galleries


nirnaor (approved):
I think that you can use some typescript here instead of integers

packages/lib/test/addMigratedOptions.spec.js
```diff
@@ -432,6 +434,7 @@ function expected() {
     layoutParams_navigationArrows_padding: 23,
     layoutParams_navigationArrows_verticalAlignment: 'ITEM_CENTER',
     layoutParams_navigationArrows_size: 23,
+    layoutParams_navigationArrows_type: 0,
```
IzaacAyelin: const

packages/blueprints/test/createBlueprint.spec.js
```diff
@@ -29,6 +29,14 @@ function getExpected() {
       options: {
         layoutParams: {
           cropRatio: 1,
+          navigationArrows: {
+            type: 0,
```
IzaacAyelin: const

yhattav (approved):

PR #689, opened 2021-12-27
Fix Layout Fixer alignment in RTL 

guysopher:


nirnaor (approved):
needs @yhattav as well

IzaacAyelin (approved):

PR #690, opened 2021-12-28
improve accessibility helper func to work on ssr

yuvalab:


packages/gallery/src/components/item/itemView.js
```diff
@@ -783,13 +782,14 @@ class ItemView extends React.Component {
 
 
   getItemAriaLabel() {
-    const { type, alt, options } = this.props;
+    const { type, calculatedAlt, options } = this.props;
     const mapTypeToLabel = {
       'dummy': '',
       'text' : extractTextItemContent(this.props.html),
```
yhattav: what happens when the text item is empty?

packages/gallery/src/components/item/itemView.js
```diff
@@ -783,13 +782,14 @@ class ItemView extends React.Component {
 
 
   getItemAriaLabel() {
```
yhattav, line 783: why is this a function in the view element(react), can this move to galleryItem / itemsHelper?
in galleryItem we could have this as part of the calculatedAlt

yhattav (changes requested):

nirnaor (approved):

yhattav (approved):

PR #726, opened 2022-01-24
enablePlayButton

IzaacAyelin:
enablePlayButton should be backward converted to "showVideoPlayButton"


yhattav (changes requested):
what is the slowArgs file?

IzaacAyelin:
@yhattav slowArgs file holds initial params for the creation of the blueprint in the test, I just changed there "showVideoPlayButton"  from true to false, the file was written in one line so that is why we see the whole file here

yhattav (approved):

PR #733, opened 2022-01-26
use boundingClientRect

IzaacAyelin:


packages/gallery/src/components/styles/gallery.scss
```diff
@@ -6,7 +6,11 @@
 *[data-hidden="true"] .pro-gallery-parent-container .gallery-item {
   background-image: none !important;
 }
-
+.pro-gallery-parent-container::after{
```
IzaacAyelin, line 9: the gallery container height collapses due to the use of the "float" property in child elements, this causes the isGalleryInViewport function to return "true" only if the top of the gallery has entered the viewport. we want it to return "true" if any part of the gallery (top or bottom) has entered the screen
using this "clear fix" solution prevents this case by maintaining the parent container height according to content

PR #761, opened 2022-03-02
use state to track scroll and check visiblity

IzaacAyelin:
**changes related scroll event:**

1. rewrite `isGalleryInViewport` relies on document API and values which is relevant only to galleries on the page (which creates a bug in lightboxes/modals)  instead use the value from the scroll event
2. keep track of the scroll top state and trigger visibility check on `componentDidUpdate` only if it was updated
3. scroll actions pass through eventsListener `GALLERY_SCROLLED` event

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -718,7 +718,6 @@ export class GalleryContainer extends React.Component {
         ...this.state.scrollPosition,
         ...eventData,
       };
-      console.log(newScrollPosition);
       this.setState({scrollPosition: newScrollPosition})
     }
   }
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -47,7 +47,10 @@ export class GalleryContainer extends React.Component {
       this.setPlayingIdxState
     );
     const initialState = {
-      scrollTop: 0,
+      scrollPosition: {
+        top: 0,
+        left: 0
+      },
       showMoreClickedAtLeastOnce: false,
       initialGalleryHeight: undefined,
       needToHandleShowMoreClick: false,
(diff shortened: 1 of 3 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -677,6 +679,14 @@ export class GalleryContainer extends React.Component {
     item?.offset && this.onGalleryScroll(item.offset);
   }
 
+  setScrollTop = (top) => {
```
nirnaor: @IzaacAyelin 
[suggestion]
if the method isn't reused elsewhere you can remove it

nirnaor (approved):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -704,6 +714,15 @@ export class GalleryContainer extends React.Component {
...
       this.props.eventsListener(eventName, eventData, event);
     }
+
+    if (eventName === GALLERY_CONSTS.events.GALLERY_SCROLLED) {
+      const { top, left } = eventData;
+      this.videoScrollHelper.trigger.SCROLL({
+        top,
+        left,
+      });
+      this.setScrollTop(top)
```
yhattav: set scroll state. the scroll state is an object of {top, left}
Functions can use either one fut in these functions we must update on what happened. not what we want to use right now.

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -677,6 +679,14 @@ export class GalleryContainer extends React.Component {
     item?.offset && this.onGalleryScroll(item.offset);
   }
 
+  setScrollTop = (top) => {
+    if (top >= 0) {
+      this.setState({
+        scrollTop: top,
```
yhattav: setState ({scroll})
where 
scroll = {top,left}

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -552,29 +552,31 @@ export class GalleryContainer extends React.Component {
...
         isInViewport,
       });
     }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState) {
    // in order to update when container is available
-   this.updateVisibility();
+   const { container } = this.props;
```
yhattav: why isnt this all part of a function called update Visibility if needed

yhattav: I dont want free floating logic in general lifecycle functions

yhattav (changes requested):

PR #765, opened 2022-03-07
Add horizontal check for getting more items when scrollless

yoavva:
… disabled


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -87,14 +87,24 @@ export class GalleryContainer extends React.Component {
     }
   }
 
+  isScrollingDisabledSituation() {
```
yhattav: isScrollingUnavailable

yhattav (changes requested):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -87,14 +87,24 @@ export class GalleryContainer extends React.Component {
     }
   }
 
+  isScrollingUnavailable() {
+    const extraPadding = 100;
+    if(this.isVerticalGallery()) {
+      // If here then the vertical scrolling is what matters
+      const { body, documentElement: html } = document;
+      const viewportHeight = window.innerHeight;
+      const height = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight);
```
nirnaor: [suggestion]
You are relying on the browser to determine the **height**. I would move these 3 lines to the `didMount` and pass the `height` down to this method. 
I prefer that browser related API's will be closer (if possible inside) the `didMount`.
If someone accidentally moves this method to the `render` phase (without knowing that it is dependant on the browser), he won't know what is wrong.

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -87,14 +87,24 @@ export class GalleryContainer extends React.Component {
     }
   }
 
+  isScrollingUnavailable() {
+    const extraPadding = 100;
```
nirnaor: [question]
How did you calculate this number?

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
yoavva: @nirnaor The `getMoreItemsIfScrollIsDisabled` function itself is called from the `didMount`, the new `isScrollingUnavailable` function is only called from `getMoreItemsIfScrollIsDisabled` --> it's called only on the didMount phase

nirnaor: I know it's called from there. I just think it's better to lift browser related dependencies up (and inject them to the method)

PR #818, opened 2022-04-17
fix-thumbnails-logic

shmuelhizmi:
https://webmaster8725.editorx.io/greenwood-main?petri_ovr=specs.pro-gallery.blockCropOnSSR:true;specs.ExcludeSiteFromSsr:true


nirnaor (approved):
I really like the fact that the change is contained in only 2 files. It shows that the code is separated correctly.

PR #819, opened 2022-04-17
fix-zindex-of-thumbnails

shmuelhizmi:
https://webmaster8725.editorx.io/greenwood-main?petri_ovr=specs.pro-gallery.blockCropOnSSR:true;specs.ExcludeSiteFromSsr:true


nirnaor (approved):
so the problem was another class on the page that was called `highlight`?

shmuelhizmi:
Yes

PR #832, opened 2022-05-08
bug simulate scroll on scrolless galleries

yuvalab:
fixing slideshow behavior - not Sliding automatically - Below the Fold https://jira.wixpress.com/browse/PHOT-3003

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -701,8 +701,8 @@ export class GalleryContainer extends React.Component {
     this.currentSlideshowViewIdx = idx;
   }
 
-  simulateScrollToItem(item) {
-    item?.offset && this.onGalleryScroll(item.offset);
+  simulateHorizontalScrollToItem(item) {
+    item?.offset && this.onGalleryScroll({left: item.offset.left});
   }
 
   eventsListener(eventName, eventData, event) {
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -701,8 +701,8 @@ export class GalleryContainer extends React.Component {
     this.currentSlideshowViewIdx = idx;
   }
 
-  simulateScrollToItem(item) {
-    item?.offset && this.onGalleryScroll(item.offset);
+  simulateHorizontalScrollToItem(item) {
+    item?.offset && this.onGalleryScroll({left: item.offset.left});
```
nirnaor, line 705: I don't understand what changed in this PR (other than rename of the method).
Before you change:
`this.onGalleryScroll` was called with `item.offSet` **(it's an object that contains `left`)**

After your change:
this.`this.onGalleryScroll` is called with an object that looks exactly the same ({`left: item.offsetLeft} **(it's an object that contains `left`)**

Unless I'm missing something, the PR just renames the function

nirnaor (changes requested):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
yuvalab: in this case i want that top won’t change in this.onGalleryScroll, i want to change only the left property. when i send all the obj the top in every scroll become 0, and this affects the scroll position.

nirnaor (approved):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
nirnaor: Understood. Are you sure it won't break anything related to vertical scroll? Because you disabled the vertical scroll behaviour by doing this

yuvalab: this func is only for scrolless galleries - and scrolless  galleries defined like this:
       scrollDirection === GALLERY_CONSTS.scrollDirection.HORIZONTAL &&
      slideAnimation !== GALLERY_CONSTS.slideAnimations.SCROLL

PR #874, opened 2022-06-26
Fix thin lines issue

guysopher:


yhattav (changes requested):
This is breaking. Why not patch our resizer (in the photography-client-lib) to disregard this and only use the const?

guysopher:
> This is breaking. Why not patch our resizer (in the photography-client-lib) to disregard this and only use the const?

Why is it breaking? It's an internally used logic that concerns only one type of source

yhattav (approved):

PR #896, opened 2022-07-18
add width to container to prevent thumbnail wraping

IzaacAyelin:


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -902,6 +902,9 @@ export class GalleryContainer extends React.Component {
         onMouseLeave={this.onMouseLeave}
         ref={(e) => (this.galleryContainerRef = e)}
         tabIndex={-1}
+        style={{
```
yhattav: what about height? why doesnt this happen when the thumbnails are horizontal? (top)

yhattav (changes requested):

packages/gallery/src/components/gallery/proGallery/galleryContainer.js
IzaacAyelin: because of CSS property called `float`. if there is not enough room for the element to "float" to the side of the container it will drop to the bottom

PR #904, opened 2022-08-04
adding option for arrows position - mouse cursor

yuvalab:


yhattav (changes requested):
the new options are missing 'MOUSE_CURSOR' (layoutsParams.ts).

yhattav (approved):

PR #924, opened 2022-08-17
feat: mouse cursor navigation logic

yuvalab:


packages/gallery/src/components/helpers/renderCursorController.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { CursorController } from './mouseCursorPosition';
+import { renderArrowButton } from './navigationArrowUtils';
```
yhattav: This is circular

yhattav (changes requested):

PR #935, opened 2022-09-05
lib - create isTouch func and use it to disabled mouse cursor feat on touch device

yuvalab:


packages/lib/src/common/utils/index.js
```diff
@@ -264,6 +265,19 @@ class Utils {
     return this.getOrPutFromCache('isMobile', _isMobile);
   }
 
+  isTouch() {
+    const _isTouch = () => {
+      const isTouchByProps = isDeviceTypeTouch();
+      const isTouchBrowser =
+        window.navigator.maxTouchPoints > 0 ||
```
yhattav: there could be no navigator on the window.

yhattav (changes requested):

yhattav (approved):

PR #941, opened 2022-09-14
Stop using stupid cropOnlyFill param

guysopher:


guysopher:
@yhattav 
We have an issue here that this fix makes the default of a slider to be FIT instead of FILL.
How can we avoid changing live sites?

PR #952, opened 2022-10-20
Do not try to infer link text from link

IzaacAyelin:
**What**
Do not try to extract a title from the provided link when link text is not given and just use link value as text

**Why**
Trying to cover the different link formats and extract a title should not be our responsibility and can be done by the user using the link `text` field
current code produces the following text for links:
`https://www.google.com => www`
`https://google.com => google`

### DIFF
--- a/packages/layouts/src/classes/galleryItem.js
+++ b/packages/layouts/src/classes/galleryItem.js
@@ -829,7 +829,7 @@ class GalleryItem {
           return 'Go to Link';
         }
       case 'web':
-        return this.linkTitleFromUrl || this.linkUrl;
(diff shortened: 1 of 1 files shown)


yhattav (changes requested):
This is breaking right?
Why do we want to change this? is it considered a bug?

IzaacAyelin:
@yhattav there is an issue when items are not provided with Organize Media (which sets the text to be the link value from the start) like From CMS or wix code without specifying text for link. so far only 1 complaint.
Screen Shot 2022-10-20 at 16 39 01

PR #984, opened 2022-11-09
version 5 beta

yhattav:


nirnaor (changes requested):
You will also need package json changes in order for the version to start bumping correctly.

yhattav:
> You will also need package json changes in order for the version to start bumping correctly.

can you explain? should I bump all versions to 5.0.0 in the different packages?

nirnaor (approved):

nirnaor (changes requested):
Just to be extra cautious I would change the external packages version to be beta also

PR #992, opened 2022-11-17
replace the workflow for v4ReleaseBranch

yhattav:
Why?
We need a v4 branch to release urgent fixes to v4 of the pro-gallery

How?
This branch should be used to release new fixes. pushing to this brach will release a version 4 patch.


nirnaor (approved):
It looks safe. You can test it and see how it publishes to make sure it's working like you want

PR #998, opened 2022-12-04
Expand Mode Preview Bug

talreu:
https://jira.wixpress.com/browse/PHOT-3295

React.Fragment (<></>) is not accepted in older versions. So if it's an older version we return a div, otherwise - React.Fragment.


nirnaor:
Please add a description of why are you doing this change and what it solves

yhattav (approved):

PR #1025, opened 2022-12-23
Update README.md

H-Huang:
Fix link in README


H-Huang:
ping @yhattav @yuvalab

yhattav:
@IzaacAyelin @nirnaor 
I think that we have a problem in the build/test flow that blocks external contributions.
If the user is not a maintainer it does not pass the deploy stage. and fails the build.
WDYT?

PR #1061, opened 2023-01-17
Media scroll helper

shmuelhizmi:


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -46,9 +46,17 @@ export class GalleryContainer extends React.Component {
     this.getIsScrollLessGallery = this.getIsScrollLessGallery.bind(this);
     this.onMouseEnter = this.onMouseEnter.bind(this);
     this.onMouseLeave = this.onMouseLeave.bind(this);
-    this.videoScrollHelper = new VideoScrollHelperWrapper(
-      this.setPlayingIdxState
-    );
+    this.mediaScrollHelper = new MediaScrollHelperWrapper([
+      {
+        getPlayTrigger: (options) =>
```
yhattav, line 51: I Like this.
But, I would even like it more if it was (later when there is also a 3d api) as a class that extends MediaScrollHelperWrapper. (galleryContainer could then load both w/o having to hold the code of the specific functions (and they would be in a specific class).

yhattav (approved):

PR #1087, opened 2023-01-29
change options of info element to use the V5 new option convention

liatv:


shmuelhizmi (approved):

packages/gallery/src/components/item/itemView.js
```diff
@@ -261,16 +261,15 @@ class ItemView extends React.Component {
   shouldShowHoverOnMobile() {
     if (utils.isMobile()) {
       const {
+        controllers_layoutParams_info_allowDescription: allowDescription,
+        controllers_layoutParams_info_allowTitle: allowTitle,
```
shmuelhizmi: remove controllers_

shmuelhizmi (changes requested):

packages/gallery/src/components/item/itemView.js
```diff
@@ -294,8 +290,7 @@ class ItemView extends React.Component {
         hoveringBehaviour !==
           GALLERY_CONSTS[
             optionsMap.behaviourParams.item.overlay.hoveringBehaviour
-          ].NEVER_SHOW &&
-        (allowDescription || allowTitle || isStoreGallery)
+          ].NEVER_SHOW
       ) {
```
liatv, line 294: @shmuelhizmi I removed allowDescription || allowTitle || isStoreGallery params from the condition, they aren't related to the core

packages/gallery/src/components/item/itemView.js
liatv: @shmuelhizmi I removed allowDescription || allowTitle || isStoreGallery params from the condition, they aren't related to the core

PR #1089, opened 2023-01-30
Add rotate arrow

yhattav:


packages/gallery/src/components/item/media/mediaItem.tsx
```diff
@@ -97,13 +98,27 @@ export default function MediaItem<T extends Record<string, any>>(
       url={isVideoPlaceholder ? videoPlaceholderUrl : props.videoUrl}
     />
   );
+  const getPlayButtonComponentByItemType = (type: string) => {
+    console.log(type);
```
shmuelhizmi: console log

shmuelhizmi: move outside function scope, this is pure

shmuelhizmi (changes requested):

PR #1092, opened 2023-02-01
add ui injection

yhattav:


packages/gallery/src/components/item/videos/videoItemWrapper.tsx
```diff
@@ -20,12 +20,17 @@ class VideoItemImplementation extends React.Component<MediaImplementationProps>
     const shouldForceLoadVideo = useTransparentPlayButtonAndForceLoadVideo(
       this.props
     );
+    console.log(this.props.showPlayButton, 'showPlayButton');
```
shmuelhizmi: hello

packages/gallery/src/context/GalleryContext.js
```diff
@@ -1,23 +1,19 @@
 import React from 'react';
 
-export let GalleryContext;
-try {
-  GalleryContext = React.createContext({});
-} catch (e) {
-  GalleryContext = null;
-}
+const GalleryContext = React.createContext({});
+console.log(
```
shmuelhizmi: :)))

packages/gallery/src/components/item/media/GalleryUI.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface GalleryUIProps {
+  size: number;
+  type: 'videoPlayButton' | 'rotateArrow';
+}
+
+export const GalleryUI = ({ type, size }: GalleryUIProps): JSX.Element => {
+  let Component;
+
+  const galleryUI = useGalleryUI();
+  if (typeof galleryUI?.[type] === 'function') {
```
liatv, line 20: why would galleryUI.type would be a function

packages/gallery/src/context/GalleryContext.js
shmuelhizmi: what about something like
```tsx
import React from "react";

export function createComponentProvider<
  T extends Record<string, React.ComponentType<any>>
>(defaultComponents: T) {
  const context = React.createContext<T>(defaultComponents);
  function OverrideUI({
    components,
    children,
  }: {
    components: Partial<T>;
    children: React.ReactNode;
  }) {
    return (
      <context.Consumer>
        {(parentComponents) => (
          <context.Provider value={{ ...parentComponents, ...components }}>
            {children}
          </context.Provider>
        )}
      </context.Consumer>
    );
  }
  const components = Object.keys(defaultComponents).reduce(
    (acc, key) => ({
        ...acc,
        [key]: (props: any) => (
            <context.Consumer>
                {(components) => {
                    const Component = components[key];
                    return <Component {...props} />;
                }}
            </context.Consumer>
        ),
    }),
    {}
    ) as T;

    return { OverrideUI, components };
}


// Usage

const { OverrideUI, components } = createComponentProvider({
    Button: ({ children }) => <button>{children}</button>,
    Input: ({ children }) => <input value={children} />,
});

const { Button, Input } = components;

export {
    OverrideUI,
    Button,
    Input,
}

const App = () => (
    <OverrideUI components={{ Button: ({ children }) => <button>{children}!</button> }}>
        <Button>Click me</Button>
        <Input>Some text</Input>
    </OverrideUI>
);
```

shmuelhizmi (changes requested):

packages/gallery/src/context/GalleryContext.js
liatv: KING

packages/gallery/src/components/item/media/GalleryUI.tsx
yhattav: galleryUI is an object of custom components

packages/gallery/src/context/GalleryContext.js
yhattav: I'll push this and then we can refactor it further. I like the fact that the components can be imported by name and not a prop (type).

PR #1115, opened 2023-03-06
add goup active index and group navigation to panel api

IzaacAyelin:


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -882,6 +882,9 @@ class SlideshowView extends React.Component {
         getActiveItemIndex: () => {
           return this.state.activeIndex;
         },
+        getActiveGroupIndex: () => {
+          return this.getCenteredItemOrGroupIdxByScroll('groups');
```
yhattav, line 886: Why would this be different than line 883?
Why isnt this a state? this calculation should not be triggered by request, but by change.

packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -892,6 +895,20 @@ class SlideshowView extends React.Component {
 
           this.props.actions.eventsListener(GALLERY_CONSTS.events.ITEM_ACTION_TRIGGERED, props, e);
         },
+        scrollToGroupIndex: (idx) => {
```
yhattav, line 898: When is this used? Please add a description on the PR.

denzelford: Adding Group Active Index and Group Navigation to the Panel API would be a fantastic enhancement! This feature could streamline navigation and provide developers with better control over grouped components[,](https://www.techbiti.com/snapdragon-profiler/) ensuring a smoother user experience and more organized UI management.

PR #1122, opened 2023-03-15
downgrade node version

IzaacAyelin:


denzelford:
To downgrade your Node.js version, you can use Node Version Manager (NVM). First, install NVM if you haven't already, then run nvm install [version] to install the desired version and nvm use [version] to switch to it. This allows easy switching between different [Node](https://topfollowapko.net/top-follow-for-ios/).js versions.

PR #1128, opened 2023-05-21
Adding tilt animation option on hover

hadasze:


liatv:
Looks good, 
lets add a numeric silder that controls the tilt angle

liatv:
also, you need to add new snapshots for the e2e (see the failing test)

PR #1132, opened 2023-05-28
Disable click actions when arrow navigation cursor is active

hadasze:


yhattav:
There is nothing related to Magnify in this code, right?

packages/gallery/src/components/helpers/mouseCursorPosition.tsx
```diff
@@ -155,16 +160,22 @@ export class MouseCursor extends React.Component<
...
+    const removeClickListener = this.context.listen.mouseClickInteractionEvent(
+      (e) => {
+        if (!this.state.mouseIn) {
+          return;
+        }
+        const [x, y] = this.state.position;
+        if (!this.props.shouldRenderAtPosition(x, y)) {
+          return;
+        }
+        if (e.type !== 'click') {
```
liatv: why here not check only mousedown || mouseup

PR #1150, opened 2023-07-26
[SEO alt text]fix alt text fallback

noam-heller1:
If the user didn't add alt text (in the media manager), then the alt text will be empty (instead of the image title/description/file name).

### DIFF
--- a/packages/gallery/src/components/item/imageItem.js
+++ b/packages/gallery/src/components/item/imageItem.js
@@ -115,7 +115,7 @@ class ImageItem extends React.Component {
 
   getImageElement() {
     const {
-      calculatedAlt,
+      alt,
       imageDimensions,
       createUrl,
       id,
@@ -227,9 +227,7 @@ class ImageItem extends React.Component {
           data-hook="gallery-item-image-img"
           data-idx={idx}
           src={src}
-          alt={
-            typeof calculatedAlt === 'string' ? calculatedAlt : 'untitled image'
-          }
+          alt={typeof alt === 'string' ? alt : 'untitled image'}
           onLoad={this.handleHighResImageLoad}
           loading={this.props.isPrerenderMode ? 'lazy' : 'eager'}
           style={{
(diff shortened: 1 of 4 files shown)


packages/gallery/src/components/item/imageItem.js
```diff
@@ -227,9 +227,7 @@ class ImageItem extends React.Component {
           data-hook="gallery-item-image-img"
           data-idx={idx}
           src={src}
-          alt={
-            typeof calculatedAlt === 'string' ? calculatedAlt : 'untitled image'
-          }
+          alt={typeof alt === 'string' ? alt : 'untitled image'}
```
liatv: 'untitled image' shouldn't be ''?

packages/gallery/src/components/item/itemView.js
```diff
@@ -779,12 +779,12 @@ class ItemView extends React.Component {
...
-    const { type, calculatedAlt, htmlContent, options } = this.props;
+    const { type, alt, htmlContent, options } = this.props;
     const mapTypeToLabel = {
       dummy: '',
       text: htmlContent,
-      video: calculatedAlt || 'Untitled video',
-      image: calculatedAlt || 'Untitled image',
+      video: alt || 'Untitled video',
+      image: alt || 'Untitled image',
     };
```
liatv, line 788: 'Untitled image' / 'Untitled video' - shouldn't it be ''?

packages/layouts/src/classes/galleryItem.js
```diff
@@ -691,22 +690,14 @@ class GalleryItem {
...
-  }
-
   get htmlContent() {
     return this.html?.replace(/<[^<>]*>/g, '').trim();
   }
 
   get alt() {
-    return this.metadata.alt || '';
+    return (
+      (utils.isMeaningfulString(this.metadata.alt) && this.metadata.alt) || ''
```
liatv, line 699: I think it should be 
(this.metadata.alt && utils.isMeaningfulString(this.metadata.alt)) || ''

liatv (approved):

packages/layouts/src/classes/galleryItem.js
noam-heller1: It will return 'true' instead of the alt

IzaacAyelin (approved):

PR #1161, opened 2023-08-27
Window wrapper mock

IzaacAyelin:
**What**
we want to use a partial window mock at the first client render
**Why**
we want the first render to use the mock values, same values used in the SSR response to prevent hydration issues

NOTE: the mock needed to be partial because we register to the window events listener when some of our components are mounted, the mocked properties on the client are dimensions and scroll

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryContainer.js
@@ -7,6 +7,7 @@ import {
   isSEOMode,
   isPreviewMode,
   isSiteMode,
+  windowWrapper,
 } from 'pro-gallery-lib';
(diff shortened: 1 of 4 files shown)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -130,6 +131,7 @@ export class GalleryContainer extends React.Component {
   }
 
   componentDidMount() {
+    windowWrapper.isGalleryMounted = true;
```
yhattav: wrap this in a function and think of the context. isGalleryMounted is the gallery's bussiness, not a windowWrapper logic. so the windowWrapper might have a flag that says useMock(or something like this) that could change to false after the gallery calls a function.

PR #1163, opened 2023-08-31
V5 migration

IzaacAyelin:


yhattav (changes requested):
Amazing.

two things that I think you should add:
1. We recommend avoiding migration functions in V5 (v4 uses this internally) as this is a performance load.
2. in V5 there is no old format. everything is flat and new. This also includes the events API or any outgoing API from the gallery - will only include the options in their new format - requiering a migration to your handler functions as well/custom components as well.

liatv (approved):

yhattav (approved):

PR #1171, opened 2023-11-22
video overlay bug fix (edit mode)

Galpittel:


packages/gallery/package.json
```diff
@@ -37,13 +37,13 @@
...
     "@vimeo/player": "2.8.2",
-    "hls.js": "^1.1.5",
+    "hls.js": "1.4.12",
     "pro-gallery-blueprints": "5.0.72",
     "pro-gallery-lib": "5.0.72",
     "pro-layouts": "5.0.72",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-player": "^2.6.2",
+    "react-player": "2.11.0",
```
liatv, line 46: isn't the change in 'MediaItem.tsx' solves the issue? why do we need to fix those packages?

liatv: @shmuelhizmi why without '^'

liatv: @shmuelhizmi why without '^'?

PR #1179, opened 2023-12-23
[BUG]  - image / container border pushes image instead of it resizing automaticly

liatv:
Fix - https://wix.monday.com/boards/4776391986/pulses/5238889014

Removing const image size and settings `box-sizing:border-box`  attribute at the image container so it can be responsive to border changes

before - 
image

after - 
image

### DIFF
--- a/packages/gallery/src/components/item/imageItem.js
+++ b/packages/gallery/src/components/item/imageItem.js
@@ -59,7 +59,7 @@ class ImageItem extends React.Component {
   }
 
   getImageContainer(imageRenderer, classNames, extraNodes) {
-    const { imageDimensions, id, actions } = this.props;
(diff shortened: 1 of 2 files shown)


yhattav:
Hey,
I understand the request here.
But, how does this work with the overlay? Why didnt you show it on the after image?
We use the layout's image size to make text layouting calculations in the overlay. if those are shortened by pure css, the calculations will go wrong.
We can make product decisions, but only after knowing that.

Again, not a bug, product decisions are needed - but you must present all the implications of this change. (also, only in v5, maybe in a minor)

liatv:
This issue needs to be solve be either changing CSS on Wix side or adding border to settings on native

PR #1184, opened 2024-02-15
fix scroll to first video on autoplay

noam-heller1:
This fix is for desktop and mobile. 
left was undefined when target.scrollLeft = 0 and target.x = undefined, so it didn't triggered the scroll.

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
@@ -59,7 +59,7 @@ export default class ScrollIndicator extends React.Component {
     this.onHorizontalScroll = (e) => {
       this.props.setGotFirstScrollIfNeeded();
       const target = e.currentTarget || e.target || e;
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/galleryScrollIndicator.js
```diff
@@ -59,7 +59,7 @@ export default class ScrollIndicator extends React.Component {
     this.onHorizontalScroll = (e) => {
       this.props.setGotFirstScrollIfNeeded();
       const target = e.currentTarget || e.target || e;
-      let left = target && (target.scrollX || target.scrollLeft || target.x);
+      let left = target && (target.scrollX || target.scrollLeft !== undefined ? target.scrollLeft : target.x);
```
liatv:
```suggestion
      let left = target && (target.scrollX || target.scrollLeft ?? target.x);
```

noam-heller1: @liatv changed both

IzaacAyelin (approved):

PR #1189, opened 2024-02-20
fix video not playing in expand mode on safari

noam-heller1:
when multiple elements have the same id, Safari selectors return the first element with that id. 

Jira ticket: https://jira.wixpress.com/browse/PG-422
Monday Item: https://wix.monday.com/boards/5762041197/pulses/6042196244


liatv (approved):
🥇

PR #1195, opened 2024-03-05
[SEO] add alt to preload images for SEO crawlers

noam-heller1:
Website crawlers (like screamingFrog) reads the first alt text and treats it as anchor text. the placeholder image had an empty alt text -> anchor text was empty. 

Best description for the problem: https://docs.google.com/document/d/1h1uKcAMsPnQ1_esjGbMUBq-NmYhwyMs6Mby7BzsD06E/edit?pli=1#heading=h.h2h41eg0msxb 

We need each link to have alt **or** anchor text, here is the result: (Yana from SEO approved it)
Screenshot 2024-03-05 at 19 21 40

### DIFF
--- a/packages/gallery/src/components/item/imageItem.js
+++ b/packages/gallery/src/components/item/imageItem.js
@@ -119,6 +119,7 @@ class ImageItem extends React.Component {
 
     const image = () => {
       const imagesComponents = [];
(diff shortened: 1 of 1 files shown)


liatv (approved):
👑

PR #1201, opened 2024-03-13
[A11y] - Switching image using the Enter key

Galpittel:
Add the ability to switch the displayed image in the gallery by pressing the Enter key on one of the thumbnails.


packages/gallery/src/components/gallery/proGallery/navigationPanel.js
```diff
@@ -96,6 +96,11 @@ class NavigationPanel extends React.Component {
                 style={itemStyle}
                 onClick={() => this.scrollToThumbnail(idx)}
                 tabIndex={0}
+                onKeyDown={(e) => {
+                  if (e.key === 'Enter') {
```
liatv, line 100: I think it's better to check with keyCode
also - we need to enter this code to V4 as well so it would be in TPA

PR #1202, opened 2024-03-14
Click on image using enter key

Galpittel:
Using a const to store the value of the 'Enter' key.

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/navigationPanel.js
+++ b/packages/gallery/src/components/gallery/proGallery/navigationPanel.js
@@ -4,7 +4,7 @@ import { GalleryUI } from '../../item/media/GalleryUI';
 
 import TextItem from '../../item/textItem.js';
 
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/navigationPanel.js
```diff
@@ -33,6 +33,7 @@ class NavigationPanel extends React.Component {
         activeIndexOffsetMemory: this.activeIndexOffsetMemory,
         prevActiveIndex: this.prevActiveIndex,
       });
+    const enterKey = 'Enter';
```
liatv:
```suggestion
    const ENTER_KEY = 'Enter';
```

liatv: Plus move it to the top of the file

liatv (approved):

PR #1207, opened 2024-04-04
Add animation duration to navigation to index

noam-heller1:
add animation (navigation) time on the navigation to item x


liatv (approved):
🥇

PR #1248, opened 2024-06-13
Blueprints dimensions changed

IzaacAyelin:


packages/blueprints/src/Blueprints.js
```diff
@@ -180,14 +180,14 @@ class Blueprints {
...
+                  existingItem.metaData.description ||
+                newItem.metaData.width !== existingItem.metaData.width ||
+                newItem.metaData.height !== existingItem.metaData.height)) ||
             (newItem.metaData &&
               newItem.metaData.type === 'text' &&
               existingItem.metaData &&
               existingItem.metaData.type === 'text' &&
-              (newItem.metaData.width !== existingItem.metaData.width ||
-                newItem.metaData.height !== existingItem.metaData.height ||
-                newItem.metaData.html !== existingItem.metaData.html ||
```
liatv, line 190: Why this was deleted?

IzaacAyelin: I moved it up for the general items check since width and height is not only relevant to text items but for all types

PR #1257, opened 2024-06-27
add the correct action check for video onclick

Dalcvi:


yhattav:
Hey @Dalcvi 
Thanks for contributing.
Can you please elaborate what this should fix?
Whats the current wrong behavior and what does this achieve?

Dalcvi:
Hey, I no longer work at wix and I dont exactly remember the problem but I think it was actually a problem on our own end with the conversion of v4 config to v5. I believe we can close this PR. Have a good day!

PR #1258, opened 2024-07-03
add a gallery navigation api event

Galpittel:


yhattav (commented):
What is this PR?
The v4 doesnt have the navigation api feature, we wrote it only in 5 AFAIK

Galpittel:
> What is this PR? The v4 doesnt have the navigation api feature, we wrote it only in 5 AFAIK

You're right, this is just a dummy PR. Closing it

PR #1262, opened 2024-07-25
Ensure the correct video player is loaded

noam-heller1:
[WIP] need to check this PR version to ensure that the correct video player is loaded before the video is played


packages/gallery/src/components/item/videos/videoItem.js
```diff
@@ -114,8 +113,22 @@ class VideoItem extends React.Component {
...
+    if (
+      (this.isHLSVideo() && !hlsPlayerLoaded) ||
+      (this.props.videoUrl &&
+        this.props.videoUrl.includes('vimeo.com') &&
+        !vimeoPlayerLoaded) ||
+      !reactPlayerLoaded
+    ) {
+      return;
+    }
+
```
liatv: insert to try

liatv (approved):

PR #1274, opened 2024-08-18
prevent listening and manipulating the anchor navigations when using a direct link

Galpittel:
stop propagation only if we are not using direct link.
stop propagation will prevent listening and manipulating the anchor navigations as some users do.
we don't want to do this when we use a direct link via a tag.


liatv (commented):
👑

PR #1290, opened 2024-09-18
send correct props when calculating gallery position

Galpittel:
When `UNSAFE_componentWillReceiveProps` in it's inner function we are calculating gallery position according to the old props (happens before props are changing).

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/slideshowView.js
+++ b/packages/gallery/src/components/gallery/proGallery/slideshowView.js
@@ -67,34 +67,34 @@ class SlideshowView extends React.Component {
     return this.state.activeIndex === 0;
   }
 
-  isScrollStart() {
-    const slideAnimation = this.props.options.behaviourParams_gallery_horizontal_slideAnimation;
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/slideshowView.js
```diff
@@ -1114,7 +1114,7 @@ class SlideshowView extends React.Component {
       GALLERY_CONSTS[optionsMap.behaviourParams.gallery.layoutDirection].RIGHT_TO_LEFT;
     const { hideLeftArrow, hideRightArrow } = this.state;
 
-    const isScrollStart = this.isScrollStart();
+    const isScrollStart = this.isScrollStart(props);
     const isFirstItem = this.isFirstItem();
```
IzaacAyelin, line 1118: Maybe cover also this function?

Galpittel: We're checking if `this.state.activeIndex === 0`, no props needed.

PR #1293, opened 2024-10-16
don't setTimeout when eventListener callback is not set

domakas:
In node environment, when setting timeout with `setTimeout(undefined, 0)`, it produces error `TypeError [ERR_INVALID_ARG_TYPE]: The "callback" argument must be of type function. Received undefined`. It seems like in this instance is always(?) undefined as `_eventsListener` which is passed always return undefined here: https://github.com/wix-incubator/pro-gallery/blob/7662ecbb93d782dbea58cccdb7d44f2615a9802f/packages/gallery/src/components/gallery/index.tsx#L59

This ussually happens in unit test. Seems like browser can handle this.

Maybe this `setTimeout` can be removed overall?


domakas:
seems like nobody cares

PR #1300, opened 2024-12-05
use options.as to determine Element in itemView

AleksMWix:


yoavtsook2806 (commented):
Hey, why do you need this option? 
What are you sending there instead of div?

yoavtsook2806 (commented):
1. Please change the prop to an object named elementsOveride with one property at the moment: {item: ${yourElement}}. this prop should not be under 'options', but as a separated prop.
2. Please add this API in another PR to master as well.
3. This is the last PR which is not a bug fix that we approve for version 4 :)

AleksMWix:
#rebuild

packages/gallery/src/components/item/itemView.js
```diff
@@ -987,8 +987,9 @@ class ItemView extends React.Component {
       options.titlePlacement !== GALLERY_CONSTS.placements.SHOW_ON_HOVER &&
       !this.hasRequiredMediaUrl;
     const itemAriaRole = this.getItemAriaRole();
+    const Element = this.props.elementsOverride.item || 'div';
```
yoavtsook2806: should add '?' for this check. if elementOverride doesn't exist this line will access undefined.
something like
    const Element = this.props?.elementsOverride.item ?? 'div';

yoavtsook2806 (approved):

PR #1311, opened 2024-12-23
Avoiding using window for SSR rendering cases

hadasze:


packages/gallery/src/components/item/videos/videoItemWrapper.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { utils, isEditMode } from 'pro-gallery-lib';
+import { utils, isEditMode, windowWrapper } from 'pro-gallery-lib';
```
IzaacAyelin: import `window` and not the `windowWrapper` itself

IzaacAyelin (approved):

PR #1337, opened 2025-01-15
[A11Y] -  Load more button accessible fix

hadasze:
making load more button accessible through tab, enter, space key presses

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/galleryView.js
+++ b/packages/gallery/src/components/gallery/proGallery/galleryView.js
@@ -131,7 +131,6 @@ class GalleryView extends React.Component {
   }
 
   onKeyPress(e) {
-    console.log('hadas onKeyPress');
     switch (e.keyCode || e.charCode) {
       case 32: // space
       case 13: // enter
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/gallery/proGallery/galleryView.js
```diff
@@ -128,6 +130,30 @@ class GalleryView extends React.Component {
     }
   }
 
+  onKeyPress(e) {
```
IzaacAyelin: it is  function specific to load more button so it should reflect that and it should also be clrear what event this is in the function name - onLoadMoreButtonKeyDown

packages/gallery/src/components/gallery/proGallery/galleryView.js
```diff
@@ -128,6 +130,30 @@ class GalleryView extends React.Component {
...
+        e.preventDefault();
+        e.stopPropagation();
+        this.props.actions.toggleLoadMoreItems();
+        return false;
+      default:
+        break;
+    }
+  }
+
+  onKeyReleased(e) {
```
IzaacAyelin: same here

IzaacAyelin (approved):

PR #1342, opened 2025-01-29
call initial video on scroll outside of constructor call

IzaacAyelin:
- Call `videoScrollHelper.updateGalleryStructure `outside of `propsTostate` since it set state in a constructor
- prevent set state on unmounted components (`VideoItemWrapper` and `VideoItem`)


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -131,6 +131,22 @@ export class GalleryContainer extends React.Component {
   }
 
   componentDidMount() {
+    const scrollHelperNewGalleryStructure = {
```
yoavtsook2806, line 136: Let's extract this to another function

yoavtsook2806 (approved):

PR #1351, opened 2025-02-03
make gallery ESM compatible

dragonofdeath:
This PR fixes node esm consumption of pro-gallery library which is required to run in modern front-end framework. The fixing esm consists of:

1. Adding explicit file extensions on every import( import ... from './file' -> import ... from './file.js')  (Mandatory file extensions)[https://nodejs.org/api/esm.html#mandatory-file-extensions]
     This could be done in too ways: changing the source or changing the build output. But as pro-gallery uses tsc directly 
     there is no possibility to modify build output. 
     
2. Dropping directory imports (`import ... from './utils'` -> `import ... from '/utils/index.js'`)
    
3. Adding 'exports' field in package json to help node find cjs & esm builds (doc)[https://nodejs.org/api/packages.html#exports]. The 'main' and 'module' fields are recognized only by bundlers, node ignores them. Using `exports` field in root package json is only way to have esm & cjs builds in one package

4. Adding package.json with 'type: module'  in esm dist folder. Node has to know the file module type before executing it and there are too posibilities:
   a. having explicit file extensions `cjs` & `mjs`. 
   b. if the file ending is `js` it will decide by looking into nearest package.json and check its `type` field. If the `type` is equal to `module` it will interpret all `js` endings in the directory as `mjs` and will fallback to `cjs` in any other case (doc)[https://nodejs.org/api/esm.html#enabling]
   
 Few other things had to be done:
 1. Its very complicated to make ts-node work with esm (its used in unit tests mainly), thus I've replaced it with tsx library. And tsx can't be loaded with mock in node-16, so the bump to node-18 had to be done.
 
2. validator build was creating mixed umd/cjs/es6 module which node doesn't support. Dropped all the transpilation with borwserify, because the library itself is provided non-minified, this should not make problems for consumers.

3. added `ts-jest-resolver` for jest specs so it would resolve files in typescript as typescript iteself does (allowing to import with 'js' extension even though the actual files in source directory have `ts` extension.


package.json
```diff
@@ -54,6 +55,7 @@
     "set-value": "^4.1.0"
   },
   "dependencies": {
-    "start-server-and-test": "^1.10.6"
+    "start-server-and-test": "^1.10.6",
+    "cheerio": "1.0.0-rc.3"
```
yoavtsook2806: Why it's not a regular version?

yoavtsook2806:
@dragonofdeath see this Falcon build:
https://falcon.dev.wixpress.com/buildConfiguration/wix_private_editor_elements_PRs/3883699?buildTab=report__Root_Falcon
It changes only the usage of the pro-gallery and seems like there are some issues with the consumption of pro-gallery.
Please make sure this PR is also green.

package.json
dragonofdeath: so the thing was here that after I added other deps this one moved to v1.0.0. And its package.json shifted so the jest started failing on importing it. So I froze this thing to its original version

dragonofdeath:
> @dragonofdeath see this Falcon build: https://falcon.dev.wixpress.com/buildConfiguration/wix_private_editor_elements_PRs/3883699?buildTab=report__Root_Falcon It changes only the usage of the pro-gallery and seems like there are some issues with the consumption of pro-gallery. Please make sure this PR is also green.

Yes, I forgot to expose the css. When package.json has 'exports' field only things that are exported might be imported by consumers. Fixed that now

yoavtsook2806:
#skipreview

PR #1357, opened 2025-02-16
[A11Y] - playing videos in hover mode only when current hover on the item

hadasze:


packages/gallery/src/components/helpers/videoScrollHelper.js
```diff
@@ -121,6 +124,17 @@ class VideoScrollHelper {
     }
   }
 
+  itemHoveredUnset(idx) {
+    if (this.videoPlay !== 'hover') return;
+    if (this.IdxExistsInVideoItems(idx)) {
+      if (this.currentPlayingIdx === idx) {
```
noam-heller1, line 130: I think this is better:
if (this.IdxExistsInVideoItems(idx) && this.currentPlayingIdx === idx) {
        this.stop(idx);
}

packages/gallery/src/components/helpers/videoScrollHelper.js
```diff
@@ -121,6 +124,17 @@ class VideoScrollHelper {
...
+        this.stop(idx);
+      }
+    } else {
```
noam-heller1: you can just remove the else

packages/gallery/src/components/item/itemView.js
```diff
@@ -106,6 +106,10 @@ class ItemView extends React.Component {
   onMouseLeave() {
     if (!utils.isMobile()) {
       this.props.actions.eventsListener(GALLERY_CONSTS.events.HOVER_SET, -1);
+      this.props.actions.eventsListener(
+        GALLERY_CONSTS.events.HOVER_UNSET,
```
noam-heller1, line 110: why is there both HOVER_SET to -1 and HOVER_UNSET? Isn't it the same?

hadasze:
#skipreview

PR #1363, opened 2025-03-03
[A11Y] - Pause Video Button

hadasze:


packages/gallery/src/components/item/itemView.js
```diff
@@ -215,7 +214,17 @@ class ItemView extends React.Component {
       { ...this.props, clickTarget },
       e
     );
-
+    if (this.props.type === 'video') {
```
yoavtsook2806: why not:
this.props.type === 'video' && this.props.idx === this.props.playingVideoIdx
in one if?

yoavtsook2806 (commented):
Looking good. left a really small comment.

packages/gallery/src/components/item/itemView.js
hadasze: fixed :)

yoavtsook2806 (approved):

PR #1385, opened 2025-03-20
[A11Y] - add the ability to play a video after pausing it

Galpittel:
[A11Y] - add the ability to play a video after pausing it

### DIFF
--- a/packages/gallery/src/components/item/itemView.js
+++ b/packages/gallery/src/components/item/itemView.js
@@ -220,7 +220,7 @@ class ItemView extends React.Component {
           { ...this.props, clickTarget },
           e
         );
-      } else if (this.props.playingVideoIdx === -1) {
+      } else {
         this.props.actions.eventsListener(
           GALLERY_CONSTS.events.VIDEO_PLAYED,
           { ...this.props, clickTarget },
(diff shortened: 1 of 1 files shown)


packages/gallery/src/components/item/itemView.js
```diff
@@ -213,15 +213,20 @@ class ItemView extends React.Component {
...
-        e
-      );
+    if (this.props.type === 'video') {
+      if (this.props.idx === this.props.playingVideoIdx) {
+        this.props.actions.eventsListener(
+          GALLERY_CONSTS.events.VIDEO_PAUSED,
+          { ...this.props, clickTarget },
+          e
+        );
+      } else if (this.props.playingVideoIdx === -1) {
```
yoavtsook2806: When is this condition true?
Only when we paused the video, or also in galleries that we play on click and it has not been played yet?

packages/gallery/src/components/item/itemView.js
```diff
@@ -213,15 +213,20 @@ class ItemView extends React.Component {
...
-      );
+    if (this.props.type === 'video') {
+      if (this.props.idx === this.props.playingVideoIdx) {
+        this.props.actions.eventsListener(
+          GALLERY_CONSTS.events.VIDEO_PAUSED,
+          { ...this.props, clickTarget },
+          e
+        );
+      } else {
+        this.props.actions.eventsListener(
```
IzaacAyelin, line 224: I think the click event should be handled in the `videoScrollHelper` event handler
we should decide there what to do with the click

PR #1396, opened 2025-04-01
[A11Y] - videos controls fix for fullscreen

hadasze:
relevant for cases this is not an external video (wix video)


hadasze:
#skipreview

PR #1431, opened 2025-05-06
Assign role="img" to images with tabIndex

zanas-wix:
Focusable elements with a `tabIndex` attribute should have an appropriate `role` assigned to them. That way screen-readers will correctly recognize what's focused. 

image

https://www.w3.org/TR/WCAG22/#name-role-value

### DIFF
--- a/packages/gallery/src/components/item/imageItem.js
+++ b/packages/gallery/src/components/item/imageItem.js
@@ -197,7 +197,6 @@ class ImageItem extends React.Component {
           data-idx={idx}
           src={src}
           alt={altText}
-          tabIndex="0"
           onLoad={this.handleHighResImageLoad}
           loading={this.props.isPrerenderMode ? 'lazy' : 'eager'}
           style={{


simsib (approved):

hadasze (commented):
first thing is that this is not the element we're focusing on when going over with tab. all focus changes should be on itemView. 
second, we do take in mind role and label, 
[label](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L769) - if the item has an alt text - this is what we should present, and if not we show the item type (image/video...) 
[role](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L924) - is where we determine what is the container role (button/link...)
[correct usage
](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L1012)please explain why the existing code is not enough and what is this special case :) .

zanas-wix:
> first thing is that this is not the element we're focusing on when going over with tab. all focus changes should be on itemView. second, we do take in mind role and label, [label](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L769) - if the item has an alt text - this is what we should present, and if not we show the item type (image/video...) [role](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L924) - is where we determine what is the container role (button/link...) [correct usage ](https://github.com/wix-incubator/pro-gallery/blob/b4aa67e9f633f53f8852e0e78e922bf763b42cb4/packages/gallery/src/components/item/itemView.js#L1012)please explain why the existing code is not enough and what is this special case :) .

Existing code has a `tabIndex` applied on an `` element. The tool we use to audit A11y is [ARC Toolkit Chrome Extension](https://chromewebstore.google.com/detail/arc-toolkit/chdkkkccnlfncngelccgbgfmjebmkmce?hl=en).

After running the tool you can see that the `img.gallery-item` element is marked as "Error":
image

After adding the `role="img"` in the DevTools and re-running the ARC Toolkit, the "Error" disappears.

> first thing is that this is not the element we're focusing on when going over with tab

If this is the case, maybe `tabIndex` should be removed from `` element then?

hadasze:
@zanas-wix good point, it could be a redundant tabIndex. have you tried removing the tabindex attribute and run arc? check everything works as expected?

zanas-wix:
@hadasze Removing `tabIndex` did the trick! :)

zanas-wix:
#rebuild

hadasze (approved):

PR #1433, opened 2025-05-08
[A11Y] - adding a11y semantics to item

hadasze:
aria-label attribute of the item uses generic role, need to specify the type. 
approved already in [V4](https://github.com/wix-incubator/pro-gallery/pull/1426/files).


hadasze:
#skipreview

PR #1438, opened 2025-05-27
Add conditional rendering for item action based on item count

frontned:
We have an a11y problem when there is only one item in slideshow view.

There is a redundant tab stop.

Screenshot 2025-05-27 at 10 51 36

My contribution makes it so that if the slideshow consists of only one item we do not render `item-action`


packages/gallery/src/components/gallery/proGallery/galleryContainer.js
```diff
@@ -891,6 +891,7 @@ export class GalleryContainer extends React.Component {
           galleryContainerId={`pro-gallery-container-${this.props.id}`}
           scrollTop={this.state?.scrollPosition?.top}
           isScrollLessGallery={this.getIsScrollLessGallery(this.state.options)}
+          hideItemAction={this.props.hideItemAction}
```
IzaacAyelin: I think `disableItemFocus` would make more sense as the item action serves only as the focus target element for the item

IzaacAyelin (approved):

PR #1445, opened 2025-06-17
render empty inner div for virtualization

noam-heller1:


packages/gallery/src/components/item/itemView.js
```diff
@@ -1024,78 +1024,103 @@ class ItemView extends React.Component {
...
-        onKeyUp={this.onContainerKeyUp}
-        onClick={this.onItemWrapperClick}
-        aria-hidden={this.props.activeIndex !== this.props.idx}
-      >
-        <div
+    let innerDiv;
+    if (this.props.type === 'dummy') {
+      innerDiv = (
+        <Element
+          className={this.getItemContainerClass()}
```
IzaacAyelin: lets put all the props in an object and then pass them to the 2 elements so we don't have to write them twice

IzaacAyelin (approved):

PR #1452, opened 2025-07-21
[Pro-gallery, lib]: Add experiment for ignoring prerender mode for invisible items

itaisinai:
## Context
Sometimes (I couldn't determine in which occasions, but seems to be very rare on strong computers), when gallery has many images (30+), we showcase only the default number of them (i.e 2 or 3).
The rest of them, will be rendered in the next seconds (Amen!).

## Trouble shoot
During the investigation, we found that where we call `getVisibleItems`, we try to return the items when they're shown by the user, and we might hiding them during the prerender mode. This could be the bug.

## What I've Done
- Added experimental features property to support experiments (also in the future).
- Added specific experiment that should ignore prerender mode.

### DIFF
--- a/packages/lib/src/common/interfaces/galleryTypes.ts
+++ b/packages/lib/src/common/interfaces/galleryTypes.ts
@@ -33,6 +33,9 @@ export interface GalleryProps {
   enableExperimentalFeatures?: boolean;
   virtualizationSettings?: VirtualizationSettings;
   shouldDisableItemFocus?: boolean;
+  experimentalFeatures?: {
+    bypassPrerenderMode?: boolean;
+  };
 }
 
 export interface GalleryState {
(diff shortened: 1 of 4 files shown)


packages/lib/src/common/interfaces/galleryTypes.ts
```diff
@@ -33,6 +33,9 @@ export interface GalleryProps {
   enableExperimentalFeatures?: boolean;
   virtualizationSettings?: VirtualizationSettings;
   shouldDisableItemFocus?: boolean;
+  experimentalFeatures?: {
+    bypassPrerenderMode?: boolean;
```
IzaacAyelin: `bypassPrerenderMode` suggest that we disable prerender mode as a whole, if that is what we wanted we could just pass `isPrerenderMode`: false, but that is not the case. the prerender mode has more to it then just the visible items functionality.
maybe give it a more specific name that suggest that we want to disable the visible items logic in prerender mode.

itaisinai: Renamed to `disableVisibleItemsOnPrerenderMode`

IzaacAyelin (approved):

PR #1455, opened 2025-07-27
Feature/flex gallery

yhattav:


IzaacAyelin:
looks like when orientation is vertical then something breaks for some of the images, there are some gaps in the items
Screenshot 2025-07-28 at 19 29 55

PR #1466, opened 2025-09-03
feat: enhance thumbnail accessibility and focus management in navigation panel

Galpittel:
Thumbnails are now "focused" when its corresponding main item is focused.
Note that there is no actual 2 focused elements at the same time, we use CSS to create a focus ring to make it looks like the thumbnail is focused too.
Solves: Selected Thumbnail Keeps Being Highlighted When Navigating Through Site with Tab

### DIFF
--- a/packages/gallery/src/components/gallery/proGallery/navigationPanel.js
+++ b/packages/gallery/src/components/gallery/proGallery/navigationPanel.js
@@ -16,6 +16,16 @@ class NavigationPanel extends React.Component {
     this.scrollToThumbnail = this.scrollToThumbnail.bind(this);
   }
 
+  state = { focusedIndex: -1 };
+
+  onThumbnailFocus = (idx) => {
+    this.setState({ focusedIndex: idx });
+  };
+
+  onThumbnailBlur = () => {
+    this.setState({ focusedIndex: -1 });
+  };
+
   scrollToThumbnail(itemIdx) {
     this.props.navigationToIdxCB(itemIdx);
   }
@@ -85,6 +95,9 @@ class NavigationPanel extends React.Component {
           {items.map(({ thumbnailItem, location, idx }) => {
             const highlighted =
               idx === activeIndex % clearedGalleryItems.length;
+            const isFocused =
+              this.state.focusedIndex === idx &&
+              this.props.settings?.isAccessible;
             const itemStyle = {
               width: thumbnailSize,
               height: thumbnailSize,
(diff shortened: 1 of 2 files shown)


packages/gallery/src/components/gallery/proGallery/navigationPanel.js
```diff
@@ -16,6 +16,16 @@ class NavigationPanel extends React.Component {
     this.scrollToThumbnail = this.scrollToThumbnail.bind(this);
   }
 
+  state = { focusedIndex: -1 };
+
+  onThumbnailFocus = (idx) => {
+    this.setState({ focusedIndex: idx });
+  };
```
itaisinai:
```suggestion
  onThumbnailFocus(idx) {
    this.setState({ focusedIndex: idx });
  }
```

packages/gallery/src/components/gallery/proGallery/navigationPanel.js
```diff
@@ -16,6 +16,16 @@ class NavigationPanel extends React.Component {
...
+
+  onThumbnailBlur = () => {
+    this.setState({ focusedIndex: -1 });
+  };
```
itaisinai:
```suggestion
  onThumbnailBlur() {
    this.setState({ focusedIndex: -1 });
  }
```

packages/gallery/src/components/gallery/proGallery/navigationPanel.js
itaisinai: Just to stick with other class methods format

packages/gallery/src/components/gallery/proGallery/navigationPanel.js
```diff
@@ -106,14 +119,18 @@ class NavigationPanel extends React.Component {
                 }
                 className={
                   'thumbnailItem' +
-                  (highlighted
-                    ? ' pro-gallery-thumbnails-highlighted pro-gallery-highlight' +
+                  (highlighted ? ' pro-gallery-highlight' : '') +
+                  (isFocused
+                    ? ' pro-gallery-thumbnails-highlighted' +
                       (utils.isMobile() ? ' pro-gallery-mobile-indicator' : '')
                     : '')
```
itaisinai: Do we have `clsx` installed here? Maybe we should use it?

itaisinai (commented):
Aluf! Minor stuff only

packages/gallery/src/components/gallery/proGallery/navigationPanel.js
Galpittel: Unfortunately we don't have it installed in this repo

IzaacAyelin (approved):

PR #1476, opened 2025-11-17
Feature: Add thumbnail ratio property

itaisinai:
## Horizontal Thumbnails
`height: thumbnailSize, width: thumbnailSize * thumbnailRatio`
### Ratio: 4:3
image

### Ratio: 3:4
Screenshot 2025-11-24 at 18 45 10

## Vertical horizontal
`width: thumbnailSize, height: thumbnailSize * thumbnailRatio`
### Ratio: 4:3
Screenshot 2025-11-24 at 18 46 52

### Ratio: 3:4
Screenshot 2025-11-24 at 18 46 14

### DIFF
--- a/packages/lib/src/core/helpers/dimensionsHelper.js
+++ b/packages/lib/src/core/helpers/dimensionsHelper.js
@@ -115,15 +115,10 @@ class DimensionsHelper {
   }
 
   getThumbnailHeightDelta() {
-    const thumbnailSize = this.options[optionsMap.layoutParams.thumbnails.size];
     switch (this.options[optionsMap.layoutParams.thumbnails.alignment]) {
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].TOP:
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].BOTTOM:
-        return (
-          thumbnailSize +
-          this.options[optionsMap.layoutParams.structure.gallerySpacing] +
-          this.options[optionsMap.layoutParams.thumbnails.marginToGallery]
-        );
+        return this.getThumbnailSize();
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].RIGHT:
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].LEFT:
         return 0;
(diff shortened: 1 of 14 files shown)


packages/lib/src/core/helpers/optionsBackwardConverter.ts
```diff
@@ -34,6 +34,7 @@ function reverseMigrateOptions(flatOptionsObject) {
     [...reversedLayoutParams].map((ele) => [...ele].reverse())
   );
   oldOptions = process_new_to_old_ThumbnailAlignment(oldOptions);
+  oldOptions = process_new_to_old_ThumbnailRatio(oldOptions);
```
IzaacAyelin: Its a new option so there shouldn't be a backward converter for this one

packages/lib/src/core/helpers/thumbnailsLogic.ts
```diff
@@ -77,7 +77,10 @@ function getThumbnailsData({
 
   const withInfiniteScroll = false; // this is not supported yet
   const thumbnailSize = options[optionsMap.layoutParams.thumbnails.size];
+  const thumbnailRatio = options[optionsMap.layoutParams.thumbnails.ratio] || 1;
+  const thumbnailHeight = thumbnailSize / thumbnailRatio;
   const thumbnailSizeWithSpacing = thumbnailSize + options[optionsMap.layoutParams.thumbnails.spacing];
+  const thumbnailHeightWithSpacing = thumbnailHeight + options[optionsMap.layoutParams.thumbnails.spacing];
```
IzaacAyelin: I think there are some product question regarding the thumbnail size resulting from the aspect ratio.
do we maintain the height/width depending on the placement of the thumbnails?
for example, if the thumbnails are on the bottom/top, shouldn't the aspect ratio affect the height of the thumbnails and the same for thumbnails on the right/left with width?
in the image you attached we see that the height is reduced for the thumbnails if the aspect ratio is on the horizontal size. I think we need to ask what should be the relationship between the thumbnails size and aspect ratio depending on the placement

itaisinai: Resolved 👍

packages/lib/src/core/helpers/dimensionsHelper.js
```diff
@@ -106,19 +106,16 @@ class DimensionsHelper {
...
-    return fixedThumbnailSize;
-  }
-
   getThumbnailHeightDelta() {
+    const thumbnailSize = this.options[optionsMap.layoutParams.thumbnails.size];
     switch (this.options[optionsMap.layoutParams.thumbnails.alignment]) {
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].TOP:
       case GALLERY_CONSTS[optionsMap.layoutParams.thumbnails.alignment].BOTTOM:
-        return this.getThumbnailSize();
+        return (
```
IzaacAyelin: refactor duplication

IzaacAyelin (approved):

PR #1481, opened 2026-01-14
Fix flickering gallery only ios

itaisinai:
Reverted this PR: https://github.com/wix-incubator/pro-gallery/pull/1465/changes
The flickers are not infinite now - once you scroll down the image resets to the first and it flickers one time, then the next image flickers one time, and then it continues play regulary.
Demo: https://wix.slack.com/files/U08LJ0KRZ60/F0A997A5H19/screenrecording_01-18-2026_19-19-55_1.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "husky": "^9.1.7",
     "lerna": "^8.2.2",
     "lint-staged": "^15.5.1",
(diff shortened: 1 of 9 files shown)


itaisinai:
#rebuild

